PR #94, opened 2015-02-11
Overlapping expectations

jaroslawk:
Hello James,

I have recently started to use your mockserver and it is a really helpful tool.
Doing the testing I realized there could be an improvement in way you resolve matching of overlapping expectations. For my needs I would prefer the expectation with most of the fields set to kick in.

Please let me know what do you think?

Best Regards, 
Jaroslaw Konczak

### DIFF
--- a/mockserver-core/src/main/java/org/mockserver/mock/MockServerMatcher.java
+++ b/mockserver-core/src/main/java/org/mockserver/mock/MockServerMatcher.java
@@ -14,6 +14,8 @@ import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.Comparator;
 import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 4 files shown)


jamesdbloom:

jaroslawk:
James let me know what do you think about this change. It is still hanging here.....

jaroslawk:
Take a look and let me know. I would love to see it closed one way or another.

jamesdbloom:
Expectation currently don't overlap as there is a strict order condition.  Expectation are matched in the order that are submitted.  If the number of matching properties is take into account the order requirement will be broken.

I have implemented a solution that doesn't complete resort all the expectations on the ```expectation_matching_specificity``` branch.  However, I'm still not sure this is a good idea because this version will break the strict order condition within matching expectations.

jamesdbloom:
I have recently added support for a ```not``` operator.  This should make it easy to have overlapping expectations and control which expectation is matched for a response (or forward or callback).  I'm going to close this pull request.

jaroslawk:
James,

We got different use cases. For me mockserver is instance serving proxy for multiple clients at the time.
For example testCase1 is asking mockserver to respond for /foo with responseBody.
At the same time testCase2 is asking to respond /foo with responseBody2.

they those are separate test cases.

For now the only option I see is to run mockserver per test case and make setup on init of the test

jamesdbloom:
Typically I would suggest if you want to run tests in parallel follow the approach specified in http://www.mock-server.com/mock_server/running_tests_in_parallel.html

Basically you need to make something unique about each test then use that when you create the expectations for each for test.

This could be a cookie as per http://www.mock-server.com/mock_server/running_tests_in_parallel.html.

If you are testing an AJAX application you could use a random query string parameter in the page making the AJAX call.  In such an example the `Referer` header of the AJAX call you include the page URL of the page making the AJAX request.  

I've used both approaches successfully before for different applications, it really just depends on how your architecture is put together.

jaroslawk:
This is exactly the case. I am using header named MDC. It contains generated in testCase id so I could get my unique response from the mockserver.

The trouble is caused by expectations which I have setup for the rest of the system - default happy path.
I do not want to pass this traffic to external service, cutting off out Integration environment out of external dependencies.

That is why I needed rule: "the most specific request wins"

for happy path I got:
/someRest service without MDC header
for test case
/someRest service with MDC set in particular test

unfortunately since general rule was submitted first - it wins and I do not have the specific for the test case response.

PR #95, opened 2015-02-12
JKS as argument

nowtv-web:
It would be useful to have a static SSL certificate (i.e. one that does not change on each start-up), as such we have made some efforts to pass in the JKS and password on the command line when starting the mockserver.

Actual contributors: @downeyfe @benlancaster @hbk619

### DIFF
--- a/mockserver-client-java/pom.xml
+++ b/mockserver-client-java/pom.xml
@@ -3,7 +3,7 @@
     <parent>
         <groupId>org.mock-server</groupId>
         <artifactId>mockserver</artifactId>
-        <version>3.9.2-SNAPSHOT</version>
+        <version>3.9.3-SNAPSHOT</version>
     </parent>
 
     <modelVersion>4.0.0</modelVersion>
--- a/mockserver-client-javascript/pom.xml
+++ b/mockserver-client-javascript/pom.xml
@@ -3,7 +3,7 @@
     <parent>
         <groupId>org.mock-server</groupId>
         <artifactId>mockserver</artifactId>
-        <version>3.9.2-SNAPSHOT</version>
+        <version>3.9.3-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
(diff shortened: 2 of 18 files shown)


nowtv-web:
Just to add, it could be that #6 negates this work.

This is still a work in progress, we would like to update the code to programmatically start up the mock server ([ClientAndProxy](https://github.com/jamesdbloom/mockserver/blob/master/mockserver-netty/src/main/java/org/mockserver/integration/ClientAndServer.java) and [ClientAndServer](https://github.com/jamesdbloom/mockserver/blob/master/mockserver-netty/src/main/java/org/mockserver/integration/ClientAndServer.java)) in order to use a pre-defined certificate, however this is kind of dependent on the status of #6 I suppose.

Thoughts?

jamesdbloom:

jamesdbloom:
I am have commited the first half of a solution that will allow a property file to be provided, either on the classpath (default) or from a fixed path (using a system property).  This property file will allow the following settings:
* JKS KeyStore File Path
* JKS KeyStore Password
* PKCS12 KeyStore File Path
* Certificate Domain Name
* Subject Alternative Name - Domain Names
* Subject Alternative Name - IP Addresses
* Toggle to trigger creating PEM files for all certificates

In addition the certificate signing has been improved so that if you import the Certificate Authority certificate (i.e. using the PEM file) into your browser or JVM then all the certificates issued will appear to be correctly issued by a known Certificate Authority.

jamesdbloom:
This pull request is now covered by the work to provide a property file to extend / change the SSL certificate generation and / or keystore management, so closing pull request.

jamesdbloom:
Documentation of the new features will following shortly on the MockServer web site.

EBichel:
The features still have a lack of documentation. Is it possible to provide a self-signed CA JKS keystore? I create one with the keytool from the JDK and have problems importing it via ConfigurationProperties.

PR #97, opened 2015-02-22
Support port 0 (JVM-assigned free port)

jekh:
This PR allows the JVM to assign a free, open port, so you can start a server like this:
```java
MockServer mockServer = new MockServer(0);
int port = mockServer.getPort(); // port will not be 0
```

### DIFF
--- a/mockserver-netty/src/main/java/org/mockserver/integration/ClientAndServer.java
+++ b/mockserver-netty/src/main/java/org/mockserver/integration/ClientAndServer.java
@@ -2,22 +2,27 @@ package org.mockserver.integration;
 
 import org.mockserver.client.server.MockServerClient;
 import org.mockserver.mockserver.MockServer;
-import org.mockserver.socket.PortFactory;
 
 /**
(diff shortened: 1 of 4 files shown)


jamesdbloom:
I like the idea of using 0 and providing this to Netty, however your implementation suffered from a small race condition that resulted in the getPort() method of MockServer sometimes returning 0 depending on how quickly the constructor got to the line which sets this.port and whether this had happened before the Thread had executed or not.

I have taken that idea and committed with your name in the comment.

jekh:
Oh, interesting. How did that manifest itself? I purposely placed the `this.port = assignedPort.get()` (line 108 in the PR) statement after the `hasStarted.get()` statement (line 103), to avoid setting this.port to an unassigned port. What was the assignedPort.get() racing with?

jamesdbloom:
You correct sorry that was my mistake.  A different error lead me to believe there was a race condition that wasn't there.  I have pulled in a big more of the code.

PR #99, opened 2015-03-07
add support for signing certificates

ganskef:
The purpose of this request is a full Certificate Authority implementation wich can sign server certificates on the fly. To get the client to trust these certificates, the root certificate will be exported. You have to register it as a trusted CA with the application or device manually. This closes CR #6, if its ready (on going).

### DIFF
--- a/mockserver-core/pom.xml
+++ b/mockserver-core/pom.xml
@@ -52,6 +52,10 @@
             <groupId>org.bouncycastle</groupId>
             <artifactId>bcprov-jdk15on</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.bouncycastle</groupId>
+            <artifactId>bcmail-jdk15on</artifactId>
+        </dependency>
 
         <!-- json (object serialisation) -->
         <dependency>
--- a/mockserver-core/src/main/java/org/mockserver/socket/KeyStoreFactory.java
+++ b/mockserver-core/src/main/java/org/mockserver/socket/KeyStoreFactory.java
@@ -1,6 +1,9 @@
 package org.mockserver.socket;
 
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
 import java.math.BigInteger;
+import java.security.Key;
 import java.security.KeyPair;
 import java.security.KeyPairGenerator;
 import java.security.KeyStore;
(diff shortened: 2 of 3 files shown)


ganskef:
Please consider to help me with the TODOs and define a public interface for this.

I would love it, if somebody help me with tests whithin this infrastructure.

jamesdbloom:

jamesdbloom:
I have pulled most of your commits so the first half of a solution is commited.  This will allow a property file to be provided, either on the classpath (default) or from a fixed path (using a system property).  This property file will allow the following settings:
* JKS KeyStore File Path
* JKS KeyStore Password
* PKCS12 KeyStore File Path
* Certificate Domain Name
* Subject Alternative Name - Domain Names
* Subject Alternative Name - IP Addresses
* Toggle to trigger creating PEM files for all certificates

In addition the certificate signing has been improved so that if you import the Certificate Authority certificate (i.e. using the PEM file) into your browser or JVM then all the certificates issued will appear to be correctly issued by a known Certificate Authority.

The remaining areas still let are, some minor tweaks with PKCS12 and more manual and automatic testing.

jamesdbloom:
This pull request is now covered by the work to provide a property file to extend / change the SSL certificate generation and / or keystore management, so closing pull request.

Documentation of the new features will following shortly on the MockServer web site.

ganskef:
Hello James,

thank you for testing and integrating this feature!

I have it in my daily use. It's a caching proxy for offline use based on 
LittleProxy. This means, its manually tested.

I'm exploring your tests to put it under test. Its not so easy for me. I 
need a SSL enabled proxy MockServer with a https server and client with 
proper certificate handling.

Regards Frank


Am 06.04.2015 um 02:27 schrieb James D Bloom:

jamesdbloom:
Please check upgrade to 3.9.4 to get a released version of the features you need.  I will also be added better SSL handling in the Proxy so that when an HTTP CONNECT request is sent I auto-generate a certificate with the correct Subject Alternative Name for that domain.  I couldn't do that for the normal MockServer because the SSL handshake happens before I receive the Host header so the only way to configure Subject Alternative Name is to allow users to provide a list of domain name and a list of IP addresses in a properties file (or via a system property that take precedence over the properties file when specified).

Please let me know if any of these features are not as you need them and I can further refine them (or submit a pull request).

PR #111, opened 2015-04-15
Update activesupport version to 4.2.0

pouyan-ghasemi:
@jamesdbloom Can we update activesupport version to the latest? I tested this locally and it worked. the previous versions are backward compatible with ```daocore```. Also can we release ```1.0.4``` version of the client after this change. Thanks a lot.


pouyan-ghasemi:
Thanks a lot @jamesdbloom

PR #119, opened 2015-04-21
Support for DELETE, HEAD, OPTIONS, TRACE HTTP methods

Unisay:
I was using DELETE in test and noticed its absent from `MockServerServlet`. I think other methods might also be useful even though use cases for them are rare.

There were no unit tests for POST and PUT (only for GET) so I didn't provide unit tests for added methods as they all delegate to `mockResponse()` like `doGet` does)


jamesbloomnektan:
I'll merge this in but I think some tests should be added.  I agree there weren't any before but I'll fix that when I get a chance.

Unisay:
Thanks!

PR #121, opened 2015-04-22
Integration test for the concurrency issue

Unisay:
Test demonstrates concurrency issue inside mock server. There are two concurrent requests:
* "Slow" request (uses Delay), causes server thread to enter synchronized method org.mockserver.mock.MockServerMatcher#handle(HttpRequest) and sleep there
* "Fast" request, whose server thread fails to enter synchronized method because "slow" thread is sleeping there so its also delayed

This bug is reproducable on both: mockserver-netty and mockserver-war (servlet deployed in Jetty)

```
java.lang.AssertionError: Fast request takes longer than expected
Expected: is a value less than <2000L>
     but: <9060L> was greater than <2000L>
```

### DIFF
--- a/mockserver-netty/src/test/java/org/mockserver/server/AbstractClientServerSharedClassloadersAndTestClasspathIntegrationTest.java
+++ b/mockserver-netty/src/test/java/org/mockserver/server/AbstractClientServerSharedClassloadersAndTestClasspathIntegrationTest.java
@@ -1,10 +1,24 @@
 package org.mockserver.server;
 
+import com.google.common.util.concurrent.Uninterruptibles;
 import org.junit.Test;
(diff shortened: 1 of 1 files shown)


jamesbloomnektan:
Good catch, I have pulled you test in and fixed the issue.  I couldn't use your exactly pull request but I added your name in the commit message.

Unisay:
Thanks James, 
when do you plan to do next maven release or do you have a snapshot repo?

jamesdbloom:
Yes please see: https://oss.sonatype.org/#nexus-search;gav~org.mock-server~~~~

In addition I am planning to release in the next 4 or 5 days as soon as I've finished the website updates and any related minor code changes.

PR #123, opened 2015-04-23
Integration test to verify interrupted request

Unisay:
Consider following use-case:
- Create a long-running request
- In case if its timed-out on client - client interrupts it
- Its still useful to verify that request reached server to test timeout handling but it doesn't work currently

BTW, this was in the log:

```
SEVERE: The web application [/mockserver] is still processing a request that has yet to finish. This is very likely to create a memory leak. You can control the time allowed for requests to finish by using the unloadDelay attribute of the standard Context implementation.
```

### DIFF
--- a/mockserver-integration-testing/src/main/java/org/mockserver/integration/server/AbstractClientServerIntegrationTest.java
+++ b/mockserver-integration-testing/src/main/java/org/mockserver/integration/server/AbstractClientServerIntegrationTest.java
@@ -3613,6 +3613,33 @@ public abstract class AbstractClientServerIntegrationTest {
     }
 
     @Test
+    public void shouldEnsureThatInterruptedRequestsAreVerifiable() throws Exception {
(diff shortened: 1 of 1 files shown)


jamesbloomnektan:
I'm happy to investigate this further but my first reaction is that this might be very difficult to fix as the request hasn't actually reached the server if the client has finished streaming a complete request.  The main issue is that it is too complex to deserialize requests when not all chunks have been received yet.  Currently MockServer only fully deserializes a requests once all chunks have been received using `io.netty.handler.codec.http.HttpObjectAggregator`.

Although, I may be wrong so I'll to dig into this more to double check.

Unisay:
@jamesbloomnektan I see your point.
But its also possible that client request fully reached the server and is suspended in delay.
In my case I want to test that timeouts are properly handled on client side. If I don't verify that request reached server it may indicate a bug in my client which caused it to hang without even reaching server.

PR #125, opened 2015-04-28
Add character set support on responses

jekh:
I ran into an issue where my tests were failing on Windows but not on Linux. After some investigation, I determined that the culprit was the HttpResponse.withBody(String) method. The method stated that it would encode strings as UTF-8, but in fact was encoding strings using the platform default charset, which is often UTF-8 in Linux but varies considerably on Windows. In my case, it was defaulting to windows-1252, which encodes some common characters such as the euro sign differently than UTF-8.

I implemented a few changes to standardize and improve the encoding of Strings. Instead of always using the platform default character set to encode Strings (the behavior of String.getBytes() without a charset parameter), this PR now allows you to specify a character set by either:
- Adding a Content-Type header with a charset. This is the preferred option, as it is what the HTTP standard requires for non-default charsets.
- Explicitly specifying the charset in the withBody() call or the StringBody() constructor. Using this option will force the charset, even if the Content-Type is absent or contradicts the specified charset.

If no character set is specified using one of the above options, the String will be encoded using ISO-8859-1, which is what the HTTP spec specifies in [section 3.7.1](http://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.7.1):
> The "charset" parameter is used with some media types to define the character set (section 3.4)  of the data. When no explicit charset parameter is provided by the sender, media subtypes of the "text" type are defined to have a default charset value of "ISO-8859-1" when received via HTTP. Data in character sets other than "ISO-8859-1" or its subsets MUST be labeled with an appropriate charset value.

This change also avoids unnecessarily Base64 encoding, then immediately unencoding, BinaryBody body types before adding them to a response body.

### DIFF
--- a/mockserver-core/src/main/java/org/mockserver/mappers/ContentTypeMapper.java
+++ b/mockserver-core/src/main/java/org/mockserver/mappers/ContentTypeMapper.java
@@ -1,11 +1,33 @@
 package org.mockserver.mappers;
 
 import com.google.common.base.Strings;
+import io.netty.handler.codec.http.HttpMessage;
+import org.mockserver.model.HttpResponse;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.servlet.http.HttpServletRequest;
+import java.nio.charset.Charset;
+import java.nio.charset.UnsupportedCharsetException;
+import java.util.List;
 
 /**
  * @author jamesdbloom
  */
 public class ContentTypeMapper {
+    private static final Logger logger = LoggerFactory.getLogger(ContentTypeMapper.class);
+
+    /**
+     * The default character set for an HTTP message, if none is specified in the Content-Type header. From the HTTP 1.1 specification
+     * section 3.7.1 (http://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.7.1):
+     * <pre>
+     *     The "charset" parameter is used with some media types to define the character set (section 3.4) of the data.
+     *     When no explicit charset parameter is provided by the sender, media subtypes of the "text" type are defined to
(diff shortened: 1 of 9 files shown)


jekh:
@jamesdbloom - would you mind re-starting the travis-ci build for this PR? Looks like the travis environment was a bit unstable today.

jamesdbloom:
I've restarted it.

jamesdbloom:
To be honest the unreliability of drone.io and travis can be quite annoying, although their free so I can't really complain.

jamesdbloom:
When I get a chance I'll take a look, merge it locally and run the build locally as I can see it doesn't look like your code is breaking the build.

jekh:
Thanks, @jamesdbloom. It looks like one of the rebuilds did actually fix the travis failure. I'm not sure what was causing the failure earlier (I wasn't able to reproduce the failure locally).

jekh:
FYI -- I just rebased on top of the latest updates, and made a few javadoc corrections.

brcolow:
:+1:

jekh:
@jamesdbloom -- just thought I'd ping you on this. Anything you'd like me to do to get this PR merge-ready? I'd be happy to add unit tests, refactor, etc.

jamesdbloom:
I am half way through merging this pull-request.  The main issue is that there was a number of missing code / tests, in particular that covered serialising and de-serialising the StringBody to JSON and back.  Once I added an integration test this highlighted that if you set an expectation using the Java API then the Charset was not sent to the server.  This is nearly fixed so it should be all fully merged in the next day or so.

jamesdbloom:
I have now pushed a these changes however some more integration tests are missing, I will add these in the next couple of days.  Once I've also fixed an issue affecting the Ruby client I will do a release.

PR #128, opened 2015-04-29
Proposed solution to Ruby client problem when verifying requests

richardPFisk:
Hi,

Please could you review my problem and proposed solution. All feedback is welcome.

# Solution
To transform the base-64 encoded body value back to a string so that verification does not error when there is a stored response.

# Detailed problem
I had a problem when using the client's ``` verify ``` method. It would error with the following stack trace:

```
[2015-04-29 13:01:40] DEBUG [MockServer::MockServerClient] URL: /retrieve. Payload: {"httpRequest"=>{"method"=>#<MockServer::Model::HTTPMethod:0x007fc6ab6e1288 @value=:POST>, "path"=>"/login"}}
[2015-04-29 13:01:40] DEBUG [MockServer::MockServerClient] Got retrieve response: 200
gems/ruby-2.0.0-p643/gems/mockserver-client-1.0.4/lib/mockserver/model/enum.rb:16:in `initialize': Supplied value: BINARY is not valid. Allowed values are: BINARY (RuntimeError)
```

client gem version=1.0.4
Docker mock-server (both in the registry and one build locally)
java -Dfile.encoding=UTF-8 -Dmockserver.logLevel=WARN -jar /opt/mockserver/mockserver-netty-3.9.1-jar-with-dependencies.jar  -serverPort 8080 -proxyPort 9090

In the context of calling ``` retrieve ```, I get the stored responses with a response body whose value is base-64 encoded. Using ```puts request``` I saw that the body.value looked suspiciously like a base-64 which it turned out to be.

Here is the code which exercises the problem:

```
require 'mockserver-client'
require "net/http"
require "uri"
require 'json'

include MockServer
include MockServer::Model::DSL

docker_host = ENV["DOCKER_HOST"]
docker_host_ip = /([0-9]{1,3}[.]){3}[0-9]{1,3}/.match(docker_host)
mock_server_port = 8080

def sendLoginRequest(mock_server_hostname_or_ip, mock_server_port)
    uri_string = "http://#{mock_server_hostname_or_ip}:#{mock_server_port}/login"
    uri = URI.parse(uri_string)
    http = Net::HTTP.new(uri.host, uri.port)
    request  = Net::HTTP::Post.new(uri.request_uri)

    request.body = "{username: 'foo', password: 'bar'}"
    response = http.request(request).to_json
end

client = MockServerClient.new(docker_host_ip, mock_server_port)
expectation = expectation do |expectation|
    expectation.request do |request|
        request.method = 'POST'
        request.path = '/login'
        request.body = exact("{username: 'foo', password: 'bar'}")
    end

    expectation.response do |response|
        response.status_code = 401
        response.headers << header('Content-Type', 'application/json; charset=utf-8')
        response.headers << header('Cache-Control', 'public, max-age=86400')
        response.body = body("{ message: 'incorrect username and password combination' }")
        response.delay = delay_by(:SECONDS, 1)
    end
end
client.register(expectation)

sendLoginRequest(docker_host_ip, mock_server_port)

client.verify(request(:POST, '/login') do |request|
    request.body = exact("{username: 'foo', password: 'bar'}")
end)
```

### DIFF
--- a/.bundle/config
+++ b/.bundle/config
@@ -1,0 +1,3 @@
+---
+BUNDLE_PATH: "./mockserver-client-ruby"
+BUNDLE_DISABLE_SHARED_GEMS: '1'
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,11 @@ before_install:
   - "./scripts/install_ca_certificate.sh"
   - "travis_retry mvn test-compile -T2C"
 
+gemfile: 
+  - "mockserver-client-ruby/Gemfile"
+
 script:
-  - "./scripts/travis_build.sh"
+  - "./scripts/travis_build.sh && cd ./mockserver-client-ruby && bundle exec rake spec"
 
 jdk:
   - oraclejdk7
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,11 +4,8 @@ before_install:
   - "./scripts/install_ca_certificate.sh"
   - "travis_retry mvn test-compile -T2C"
 
-gemfile: 
-  - "mockserver-client-ruby/Gemfile"
-
 script:
-  - "./scripts/travis_build.sh && cd ./mockserver-client-ruby && bundle exec rake spec"
+  - "./scripts/travis_build.sh"
 
 jdk:
   - oraclejdk7
--- a/mockserver-client-ruby/lib/mockserver/model/body.rb
+++ b/mockserver-client-ruby/lib/mockserver/model/body.rb
@@ -11,7 +11,7 @@ module MockServer::Model
   # An enum for body type
   class BodyType < SymbolizedEnum
     def allowed_values
-      [:STRING, :REGEX, :XPATH, :PARAMETERS]
+      [:STRING, :REGEX, :XPATH, :PARAMETERS, :BINARY]
     end
   end
 
--- a/mockserver-client-ruby/lib/mockserver/model/request.rb
+++ b/mockserver-client-ruby/lib/mockserver/model/request.rb
@@ -3,7 +3,7 @@ require 'hashie'
 require_relative './parameter'
 require_relative './body'
 require_relative './enum'
-
+require "base64"
 #
 # A class to model a request in an expectation.
 # @author:: Nayyara Samuel (mailto: nayyara.samuel@opower.com)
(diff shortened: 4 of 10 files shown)


jamesdbloom:
Thanks, I'm very backed up now but I'll try to take a look at this in the next week or so.

jamesdbloom:
I am ready to merge this however I can't seem to merge this using github web tool and when I do it from the command line it marks all the commit against my name.  Do you know why the web tool is not allowing me to merge your changes?

jamesdbloom:
I also have some review comments as follows:

Can you uncomment `enum.rb:16` instead I created a simpler version of this method in the `Boolean` class as follows:

```rb
class Boolean < Enum
  def allowed_values
    [true, false]
  end

  def !
    !@value
  end

  def initialize(supplied_value)
    @value = pre_process_value(supplied_value)
  end
end
```

In addition can you revert the changes to .travis.yml as the maven build is already supposed to run the ruby tests.  However there was a bug in the way maven was configured so please add the `spec` argument to `exec-maven-plugin` plugin configuration in `mockserver-client-ruby/pom.xml` as follows:

```xml
<execution>
    <id>ruby_make</id>
    <phase>prepare-package</phase>
    <goals>
        <goal>exec</goal>
    </goals>
    <configuration>
        <skip>${skipRubyBuild}</skip>
        <executable>bundle</executable>
        <arguments>
            <argument>exec</argument>
            <argument>rake</argument>
            <argument>build</argument>
            <argument>spec</argument>
        </arguments>
        <environmentVariables>
            <BUNDLE_GEMFILE>${basedir}/Gemfile</BUNDLE_GEMFILE>
        </environmentVariables>
    </configuration>
</execution>
```

Then to run just the `mockserver-ruby-client` (as it runs on Travis) just:

```bash
cd ./mockserver-client-ruby && mvn clean install
```

If you can make these changes hopefully I can use the github web tool for merging and then your commit will appear under your name.  If you're not bothered then I'm also happy to merge using the command line tool and add your name into the comment.

richardfisk:
Thanks for the feedback. I have made the suggested changes. 

Please I note I have rebased so that the merge is clean (this removed the comment enum.rb:16 in an earlier commit).

jamesdbloom:
If you can't get the Travis build to work don't worry as I can run the build locally, the Travis build for pull requests seems flaky, if it fails the first time I re-run it but if it fails again for something unrelated to your change I'll merge if it passes on my local machine.

Also I'll be out of the country on business until Thursday so I may not get a chance to look at this until end of the week.

PR #161, opened 2015-07-31
Save expectations to a file when server exists and load them at start.

joemeszaros:
You can start mockserver with a new argument: -databaseFile.  It specifies the file name, where mockserver saves the expectations as JSON when exits, and from which the expections are loaded at startup. It was a little bit inconvenient to always reconfigure mockserver with a remote client, because my modifications were lost after the mockserver process exited.

### DIFF
--- a/mockserver-core/src/main/java/org/mockserver/mock/MockServerMatcher.java
+++ b/mockserver-core/src/main/java/org/mockserver/mock/MockServerMatcher.java
@@ -25,8 +25,16 @@ import java.util.regex.Pattern;
  */
 public class MockServerMatcher extends ObjectWithReflectiveEqualsHashCodeToString {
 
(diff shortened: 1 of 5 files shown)


joemeszaros:
The travis build fails with error message : Failed to deploy artifacts: ... Failed to transfer file: https://oss.sonatype.org/content/repositories/snapshots/org/mock-server/mockserver/3.9.18-SNAPSHOT/mockserver-3.9.18-20150731.075602-21.pom. Return code is: 401, Reason Phrase: Unauthorized.

I see keytool related error messages in the log, which can cause the 401 return code.

@jamesdbloom  Could you please take a look at it? As far as I see, other builds failed as well.

zhangyanwei:
I can't find this argument, maybe it has been removed. (but I also checked the commit history,  not found it :( )

PR #168, opened 2015-08-27
FIX: MockServerHandler dropped BlockingOperationException in some cases

joemeszaros:
### FIX : Eliminating awaitUninterruptibly() call from the MockServerHandler class, causing BlockingOperationException in some cases to prevent dead locks.

According to the Java doc of Netty's [ChannelHandlerFuture](http://netty.io/4.0/api/io/netty/channel/ChannelFuture.html) class:

*Do not call await() inside ChannelHandler*

*The event handler methods in ChannelHandler is usually called by an I/O thread. If await() is called by an event handler method, which is called by the I/O thread, the I/O operation it is waiting for might never be complete because await() can block the I/O operation it is waiting for, which is a dead lock.*

```java
 // BAD - NEVER DO THIS
 @Override
 public void channelRead(ChannelHandlerContext ctx, GoodByeMessage msg) {
     ChannelFuture future = ctx.channel().close();
     future.awaitUninterruptibly();
     // Perform post-closure operation
     // ...
 }
```

```java
 // GOOD
 @Override
 public void channelRead(ChannelHandlerContext ctx,  GoodByeMessage msg) {
     ChannelFuture future = ctx.channel().close();
     future.addListener(new ChannelFutureListener() {
         public void operationComplete(ChannelFuture future) {
             // Perform post-closure operation
             // ...
         }
     });
 }
```
*In spite of the disadvantages mentioned above, there are certainly the cases where it is more convenient to call await(). In such a case, please make sure you do not call await() in an I/O thread. Otherwise, BlockingOperationException will be raised to prevent a dead lock.*

I removed the awaitUninterruptibly() call, because `ctx.writeAndFlush(response).addListener(ChannelFutureListener.CLOSE)` is enough.

### DIFF
--- a/mockserver-netty/src/main/java/org/mockserver/mockserver/MockServerHandler.java
+++ b/mockserver-netty/src/main/java/org/mockserver/mockserver/MockServerHandler.java
@@ -240,20 +240,21 @@ public class MockServerHandler extends SimpleChannelInboundHandler<HttpRequest> 
 
     private void writeAndCloseSocket(ChannelHandlerContext ctx, HttpRequest request, HttpResponse response) {
         ConnectionOptions connectionOptions = response.getConnectionOptions();
-        if (connectionOptions != null && connectionOptions.getCloseSocket() != null) {
-            if (connectionOptions.getCloseSocket()) {
-                ctx.writeAndFlush(response).addListener(ChannelFutureListener.CLOSE).awaitUninterruptibly();
-                ctx.close();
-            } else {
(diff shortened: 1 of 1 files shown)


jamesdbloom:
I agree with you fix, however, your rewriting of the `if` block is not correct and causes some unit tests to fail.  I have applied your change in a separate commit and referred to you.

joemeszaros:
Thanks for the update

PR #176, opened 2015-09-15
Enforce TLS 1.2 if available, since it's not default up to Java 8.

ganskef:
Problem: Some servers like https://forums.freebsd.org refuse TLSv1 connections.

Solution: Enforce TLSv1.2 if available, since it's not default up to Java 8. Use TLSv1 to fall back for Java 6.

Background: Java 7 disables TLS 1.1 and 1.2 for clients. From Java Cryptography Architecture Oracle Providers Documentation: Although SunJSSE in the Java SE 7 release supports TLS 1.1 and TLS 1.2, neither version is enabled by default for client connections. Some servers do not implement forward compatibility correctly and refuse to talk to TLS 1.1 or TLS 1.2 clients. For interoperability, SunJSSE does not enable TLS 1.1 or TLS 1.2 by default for client connections.

Drawback: It's totally obsolete with Java 8, but we're supporting Java 6 at the moment. This is okay, since it is a little overhead, I think.

### DIFF
--- a/mockserver-core/src/main/java/org/mockserver/socket/SSLFactory.java
+++ b/mockserver-core/src/main/java/org/mockserver/socket/SSLFactory.java
@@ -1,6 +1,7 @@
 package org.mockserver.socket;
 
 import io.netty.handler.ssl.util.InsecureTrustManagerFactory;
+
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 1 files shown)


jamesdbloom:
Sorry for the delay in merging but I have now fixed the issue with Travis failing all PR builds.  But this PR was too early for that fix so I had to do it manually and run the build locally.

PR #211, opened 2016-01-26
Ability to add expectations to the beginning of list, ability to delete expectations by path

SHaaD94:
I have changed order of adding new expectations which <b>Times</b> aren't Unlimited. This feature is used to create temporary expectations for limited amount of requests then temporary expectations with same requests already been set.

Second feature is used to delete expectations (i've found existing method <b>Clear</b> uncomfortable to use, because exact json of expectation is required).

### DIFF
--- a/mockserver-core/src/main/java/org/mockserver/mock/MockServerMatcher.java
+++ b/mockserver-core/src/main/java/org/mockserver/mock/MockServerMatcher.java
@@ -13,6 +13,7 @@ import org.mockserver.model.ObjectWithReflectiveEqualsHashCodeToString;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.servlet.http.HttpServletRequest;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


jamesdbloom:
deleting expectations by path is already supported using the clear functionality, this matches on any part of an expectation matcher to clear only those expectations that match against the provided request matcher, for more details see: http://www.mock-server.com/mock_server/clearing_and_resetting.html

In addition by adding expectations to the front of the list the newer expectations get matched first, but only if they are not unlimited.  I think this may be a little confusing because the matching of expectations will occur in opposite order to the order the expectations are added, unless they are unlimited in which case it will occur in the order they are added.  What was the use case behind needing this change?

boqu:
Clearing expectations by path method regardless http method, queries etc is a handy for the case when I don't have exact requests details I want to remove except the path.

jamesdbloom:
Closing this PR due to the issues discussed above and the fact that the code can not be merged.

PR #215, opened 2016-02-03
Ensure ca/client cert serial number positive

jnormington:
- This caused bad certification validation errors with go 1.4
   do to adhering to the RFC3280 where the serial number should
   always be non-negative.
 - This change ensures that we pass Integer.MAX_VALUE to Random#nextInt(val)
   so that a positive number from 0 to val is generated as per the java docs


jnormington:
@jamesdbloom - we also need to regenerate new ca-cert and generate a new standalone jar for public consumption - but I haven't done this part of the PR - leaving that part to you please.

Issue #228, opened 2016-03-02
Ruby client does not accept PATCH (Or custom) HTTP methods


joseprl89:
The mockserver-client library for Ruby is currently not allowing to provide PATCH (or any HTTP verb outside of GET POST PUT and DELETE).

The error message raised when trying to use a PATCH verb is:

> Supplied value: PATCH is not valid. Allowed values are: [:GET, :POST, :PUT, :DELETE] (RuntimeError)

I think I'll be able to work around that by embedding the code for now in my ruby workspace and edit the files, however it would be preferable to have a fix. Could you add this to your roadmap?

joseprl89:
I'll start a pull request to include the PATCH verb (which is the one I need), however I think you should plan on accepting any verb, since non standard verbs could also be used.

jamesdbloom:
I am actually planning on re-writting the entire ruby client to bring it more inline with the way other client's are written and to significantly simplify it, for example by not restricting specific values for HTTP methods when this is not required, etc.  I am therefore going to close this ticket as the immediate support for PATCH has been provided in the PR and the more flexible will be introduced once I re-write the ruby client.

joseprl89:
@jamesdbloom when are you planning to release this code? I'd like to swap back to importing a gem rather than keeping a copy of the code in my repo.

PR #229, opened 2016-03-02
Added PATCH to the allowed HTTPMethod values.

joseprl89:
Partially solves #228


joseprl89:
@jamesdbloom do you have an ETA on when would this code be released via a gem?

PR #231, opened 2016-03-04
Fix java command arguments in docker run script

mishako:
~~Since commit 905a71d the COMMAND_LINE_OPTS variable doesn't have a
leading space. This results in an error: "Could not find or load main class
org.mockserver.cli.Main-Djava.library.path=.libtcnative-1.so.0.1.27".~~

Java system property java.library.path should not be a part of
COMMAND_LINE_OPTS string, because system properties must be specified
before the main class name (see java command usage).

### DIFF
--- a/docker/run_mockserver.sh
+++ b/docker/run_mockserver.sh
@@ -96,7 +96,7 @@ fi
 
 validateArgument $LOG_LEVEL "OFF ERROR WARN INFO DEBUG TRACE ALL." "Invalid value '$LOG_LEVEL' for 'logLevel'"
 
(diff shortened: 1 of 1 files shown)


mishako:
@jamesdbloom

pauvos:
+1

icohen:
+1

mishako:
Turns out adding a space doesn't fix the underlying issue. I updated the commit. With this changes I can actually run the docker image.

icohen:
verified 210ed2f works. Thanks @mishako for the fix!

PR #233, opened 2016-03-07
Added the ability to get Expectations back from retrieve for a proxy

brucedeen:
**Usage**: mockserver:1090/retrieve?type=expectations

This will pull ALL httprequest/httpresponse combinations back for a matching httprequest (or all if httprequest is null).  This is a little different then what dumpToLog was doing, but is going to be used to allow for retrieval of data from a docker container without having to print anything to the logs.

Our use case for this is fairly simple.

As a developer I would like the proxy configured mockserver to provide me every request/response combination I have made to validate an integration test has not changed, thereby allowing me to understand when contracts between services have evolved.

The following is an example of JSON serialized output based upon the unit test
`[
    {
        "httpRequest": {
            "path": "some_path"
        },
        "httpResponse": {
            "statusCode": 200,
            "body": "some_body"
        },
        "times": {
            "remainingTimes": 1,
            "unlimited": false
        },
        "timeToLive": {
            "unlimited": true
        }
    },
    {
        "httpRequest": {
            "path": "some_path"
        },
        "httpResponse": {
            "statusCode": 200,
            "body": "some_body"
        },
        "times": {
            "remainingTimes": 1,
            "unlimited": false
        },
        "timeToLive": {
            "unlimited": true
        }
    },
    {
        "httpRequest": {
            "path": "some_path"
        },
        "httpResponse": {
            "statusCode": 404
        },
        "times": {
            "remainingTimes": 1,
            "unlimited": false
        },
        "timeToLive": {
            "unlimited": true
        }
    },
    {
        "httpRequest": {
            "path": "some_path"
        },
        "httpResponse": {
            "statusCode": 404
        },
        "times": {
            "remainingTimes": 1,
            "unlimited": false
        },
        "timeToLive": {
            "unlimited": true
        }
    },
    {
        "httpRequest": {
            "path": "some_other_path"
        },
        "httpResponse": {
            "statusCode": 200,
            "body": "some_other_body"
        },
        "times": {
            "remainingTimes": 1,
            "unlimited": false
        },
        "timeToLive": {
            "unlimited": true
        }
    }
]`

### DIFF
--- a/mockserver-core/src/main/java/org/mockserver/filters/RequestResponseLogFilter.java
+++ b/mockserver-core/src/main/java/org/mockserver/filters/RequestResponseLogFilter.java
@@ -1,10 +1,8 @@
 package org.mockserver.filters;
 
 import org.mockserver.client.serialization.ExpectationSerializer;
-import org.mockserver.client.serialization.HttpRequestSerializer;
 import org.mockserver.client.serialization.java.ExpectationToJavaSerializer;
 import org.mockserver.collections.CircularMultiMap;
-import org.mockserver.logging.LogFormatter;
 import org.mockserver.matchers.HttpRequestMatcher;
 import org.mockserver.matchers.MatcherBuilder;
 import org.mockserver.matchers.TimeToLive;
@@ -15,10 +13,7 @@ import org.mockserver.model.HttpResponse;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.ArrayList;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 import static org.mockserver.model.HttpResponse.notFoundResponse;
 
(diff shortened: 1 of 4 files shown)


jamesbloomnektan:
This is an interesting idea, I was intending on adding a simple client to view and update the proxy and MockServer. However, I've recently been distracted on other projects, those are quiet now so I'm catching up and will look at this very soon.

jamesdbloom:
I'm closing this PR because in addition to the Proxy client being built.

It is deliberate that expectations are not returned by the proxy.  This is because of the anti-pattern of record and replay for recording proxies such as betamax.  The key issue is that if test mocks are created like that the mock data becomes over complex and brittle.  These types of mock are typically hard to manage and often shared across multiple tests.  For example if one or more tests are created using record and replay mocks if the data mocks need to be re-recorded (i.e. due to a new field being added to a response) then the test typically has to be complete re-written.

The proxy aims to avoid the record - replay anti-pattern by providing information about what request where made so expectations can be handle crafted easily.

PR #238, opened 2016-04-09
Convert the request body correctly when it is a String

jnormington:
so hashie coercion doesnt error when it retrieves and the requests have body content.

As currently if it retrieves a request with body content it errors out with the following;

```
Cannot coerce property :body from String to MockServer::Model::Body: undefined method `each_pair' for "Hello this is a message":String
```

It appears that the mockserver java core doesn't return the type of body and this appears to be implied by this java spec - https://github.com/jamesdbloom/mockserver/blob/master/mockserver-core/src/test/java/org/mockserver/model/HttpRequestTest.java#L96-L114

My gut feeling tells me it ideally should returned from the overridden toString methods in all the classes extending Body like StringBody class below

```
https://github.com/jamesdbloom/mockserver/blob/master/mockserver-core/src/main/java/org/mockserver/model/StringBody.java#L65-L68
```

### DIFF
--- a/mockserver-client-ruby/lib/mockserver/model/request.rb
+++ b/mockserver-client-ruby/lib/mockserver/model/request.rb
@@ -74,6 +74,12 @@ module MockServer::Model
     end
 
     def request_from_json(payload)
+      body = payload['body']
+
+      if body && body.is_a?(String)
+        payload.merge!('body' => { 'type' => :STRING, 'value' => body })
+      end
+
       request = Request.new(symbolize_keys(payload))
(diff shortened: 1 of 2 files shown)


jnormington:
@jamesdbloom on second thoughts this doesn't seem write if a request body is of type binary then this would break that (not that it is working currently) - so I guess the mockserver request response should also include the bodyType ?

jamesbloomnektan:
When the body is a binary it should include the bodyType, only bodies which are strings support the shorthand of no bodyType.

Issue #120, opened 2015-04-22
Support method / object callbacks instead of class name


JanGurda:
Thanks for nice piece of software.
 I have just started using MockServer I love it however I feel I miss one feature. While mocking I'm able to specify HttpCallback which will be executed on server invocation. However HttpCallback may contain only class name. For some use cases I would like something more dynamic. I wonder if you considered allowing to put instance rather than class into HttpCallback so user could define behavior more dynamically.

That would be more similar to Mockito's when(something).thenAnswer(code_to_execute_here)

What do you think?

jamesbloomnektan:
The problem with this is that the expectation is serialised to the server using JSON.  Typically the server runs as a separate process and so has no access to local objects.  So currently there is no way to support executing an instance object.

What you can do as a work around is make sure the MockServer is running in the same JVM by using the API such as ServerAndClient then use static methods.  However, this is pretty much a hack that I don't recommend as you would have no test separation.

I'm going to close this ticket as it is not possible due to limitation in Java.

jamesbloomnektan:
However I could implement this using web sockets, so I'll re-open this ticket and implement it when I get a chance.

JanGurda:
You are right. I missed the fact that everything goes through the wire even when we use junit Rule and server runs inside the same JVM process. Websockets is nice solution for that. We simply need a way to invoke code on MockServerClient's side and pass created response back to server. I see also other option here (very similar to websockets) - expose some kind of server endpoint on client side (could be HTTP server). 

When API user records behavior (creates expectations) with instance of given interface (similar to ExpectationCallback) client could assign unique identifier and map this unique identifier to that expectation instance. That unique id could be later passed to server and then server finds that should invoke expectation it calls client giving id and gets HttpResponse back. This is how I see that now.

Thank you for quick response.

jamesbloomnektan:
For now I think web sockets is the best approach as it works really nicely in Netty which is used for both the client and the server.  I've previously implemented web sockets with Netty before and they provide a really solid and simple implementation.

mhurne:
When using MockServer via the JUnit MockServerRule, is everything running in the same JVM such that passing a ExpectationCallback instance would be possible? In Spock tests, this would be quite powerful, as a Closure could be used to implement the callback...

jamesdbloom:
[...]

PR #258, opened 2016-06-28
Creating expectations with mock responses and http web hooks

vmironichev:
This feature gives possibility to create optionally one or more web hooks for specific expectation along with prepared http response, nodejs expectation configuration example:

``` javascript
var httpMockResponse = {
    'httpRequest': httpRequestExpectation,
    'httpResponse': {
        'statusCode': requestConfig.http_response_code,
        'body': JSON.stringify(responseBody),
    },
    'httpWebHook': {
       'httpWebHookConfig': 'endpoints': [
           {
                'host': 'localhost',
                'port': '4567',
                'path': '/',
                'payload': JSON.stringify('{"name":"test_value"}'),
                'executionDelay': 5000,
                'executionDelayTimeUnit': 'MILLISECONDS'
           }
        ]
    }
}

mockServerClient("localhost", SERVER_PORT).mockAnyResponse(httpMockResponse);
```

### DIFF
--- a/mockserver-core/src/main/java/org/mockserver/client/serialization/java/ExpectationToJavaSerializer.java
+++ b/mockserver-core/src/main/java/org/mockserver/client/serialization/java/ExpectationToJavaSerializer.java
@@ -39,6 +39,11 @@ public class ExpectationToJavaSerializer implements ToJavaSerializer<Expectation
                 output.append(new HttpCallbackToJavaSerializer().serializeAsJava(numberOfSpacesToIndent + 1, expectation.getHttpCallback()));
                 appendNewLineAndIndent(numberOfSpacesToIndent * INDENT_SIZE, output).append(")");
             }
+            if (expectation.getHttpWebHook() != null) {
+                appendNewLineAndIndent(numberOfSpacesToIndent * INDENT_SIZE, output).append(".webHook(");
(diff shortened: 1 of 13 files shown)


vmironichev:
Hi @jamesdbloom, please let me know if this feature is useful and let me know if it needs some other changes except tests) (can add them if this feature makes sense to for project)

jamesdbloom:
This functionality, as I understand, already exists using the forward action.  The forward action allows a request to be forwarded to another endpoint.  

In addition #120 contains a description of the dynamic method callbacks that allow local methods (i.e. closure) to be executed to generate the response.  If a local method / closure is called then the code executed in the closure could both return the response and call your webhook endpoint.

I'm not keen on merging this PR as it doesn't feel like it is consistent with the core feature set of MockServer.  Can you please explain the use cases when this would be useful?

vmironichev:
This feature is a bit different than http forward, since it sends to specific endpoint defined in configuration payload that is different from original request payload that was passed. This is completelly configuration driven and allows sending custom responses for a given requests within configured delay. Let me briefly describe a use case which I completed in this feature:

1. Mock server received async request
2. Reuest should be processed within certain time (executionDelay)
3. After request is processed (for example payment settled, which usually takes some hours to be processed))
4. Notify with via web hook (specific endpoint from configuration) with information (payload) whether original request was processed successfully or not.

Thanks,
Valeriy

PR #264, opened 2016-07-21
[WIP] Allow explicitly specifying protocol and/or base path

vyo:
We have a setup necessitating two additional configuration options:
1. Being able to force HTTPS
2. Specifiying a base path independent from an app context

Consider this a basis for discussion - I may split this up into two distinct requests if required and will squash/rebase as necessary .

I tried keeping it as general as possible, as well as not messing with existing code paths.

### DIFF
--- a/mockserver-client-java/src/main/java/org/mockserver/client/AbstractClient.java
+++ b/mockserver-client-java/src/main/java/org/mockserver/client/AbstractClient.java
@@ -21,8 +21,10 @@ public abstract class AbstractClient {
 
     protected final Logger logger = LoggerFactory.getLogger(this.getClass());
 
+    protected final HttpRequest.Protocol protocol;
     protected final String host;
     protected final int port;
+    protected final String basePath;
     protected final String contextPath;
     protected NettyHttpClient nettyHttpClient = new NettyHttpClient();
     protected HttpRequestSerializer httpRequestSerializer = new HttpRequestSerializer();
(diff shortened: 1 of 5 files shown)


jamesdbloom:
Can you please explain why you need a base path independent from an app context?  Why can't that be handled as a variable in your test code for example.

I can understand the requirement for restricting to SSL but I would argue the best way to achieve this would be by added this as a matcher in the request matcher so that MockServer would only return a response if the request was made by SSL.  This sort of approach fits much better with the architect and approach taken for MockServer and would make this feature easily available across all clients and mechanisms of launching MockServer.

vyo:
Aloha, thanks for checking back! We have the mockserver running behind a reverse proxy - specifying the addtional 'proxy-induced' segment before the actual call would lead to requests not being correctly matched.
I do concede that this can be remedied on the caller's side but that would make us lose the consistency between calling the mockserver and the real thing.

The forcing of SSL was necessitated by bug in the same proxy, leading to the auto-HTTPS feature not picking up cleanly. As an additional option I think it would not hurt and other people even might get other uses out of it.

Cheers!

jamesdbloom:
On the auto SSL option I would be happy to add this as a system property configured using `-D...` or programatically using `org.mockserver.configuration.ConfigurationProperties`.  In addition I don't think the way you have specified the protocol will work because you only specifying it in the client that it used to setup expectations in the MockServer and by the time MockServer can receive requests from the client it is far too late to be configuring protocol support.  All the protocol support is setup before MockServer receives any request even a request from the client.

Can you please explain in more detail why you need a base path because I don't understand what the exact use case is with the reverse proxy.  I think there may be an existing way to do what you need but I'm not sure I really understand your problem.

vyo:
Generally I had to modify both server and client to make things work.

Let me shortly lay out the details as to why the base path option was required in our context.

We run mockserver WAR-packaged in a non-root Tomcat context, behind a reverse proxy server as follows:
`proxy:443/tomcat-route/mockserver-context/${app-resources}`
Configuring the context client-side as '/tomcat-route/mockserver-context' led to the server not properly matching the request; after all, '/tomcat-route/mockserver-context' does not match '/mockserver-context' as resolved by the mockserver itself.

Hope that makes it clear. If I find the time I can create a Docker showcase.


TL;DR: Reverse proxy + Tomcat deployment messes up resource matching.

jamesdbloom:
Have you tried using a regex to allow the path to be match without or without the base path?  That way you can create request matchers that are resilient to the base path being there or not being there.  I'm not sure if that will work in your situation but I think it should give you what you need?

vyo:
That sounds like a nifty idea, I'll give it a shot as soon as I have the time to!
I'll let you know if it fits the bill.

Cheers!

PR #273, opened 2016-08-04
Allow run_mockserver.sh to accept subject alt names parameter

bjhartin:
This is handy if you're running the docker image and need the generated certs to have specific subject alternate domain names.

The need arose because we use docker-compose to spin up a mock-server, e.g.

``` yml
mock-server:
  image: jamesdbloom/mockserver
  command: "/opt/mockserver/run_mockserver.sh -logLevel DEBUG -serverPort 80,443
    proxyPort 1090 -sslAltDomains www.domain1.com, www.domain2.com"
```

### DIFF
--- a/docker/run_mockserver.sh
+++ b/docker/run_mockserver.sh
@@ -3,7 +3,7 @@
 set -e
 
 function showUsage {
-    echo >&2 "   run_mockserver.sh [-logLevel <level>] [-serverPort <port>] [-proxyPort <port>] [-proxyRemotePort <port>] [-proxyRemoteHost <hostname>]"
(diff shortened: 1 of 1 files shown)


jamesdbloom:
I would prefer to change this PR to allow pass an abstract string to the JVM for generic system preferences.  This could be used for example to update JVM setting such as memory or any generic MockServer settings such as the default subject alternative names or CORS settings.

If you willing to update the PR and submit it then that would be good, if not I'll make that change when available, hopefully the next couple of weeks.

jamesdbloom:
I'm now closing this PR as the functionality can be achieve by using the `genericJVMOptions` option with the latest docker container.

PR #285, opened 2016-10-02
remove outdated Xerces dependency and fix XML pretty print issue

nitram509:
Hi,

first of all, thank you for this great library, which I use to write HTTP client tests.

During development, I've discovered some issues with Xerces being in classpath *1)
and registering it's parsers. Modern JRE XML implementations are way more up-to-date.
Thus I removed the Xerces dependency and just using standard JRE features for pretty printing an XML.
As a result, the tests needed to be adopted, because now they are correctly pretty printed, including NewLine character.

*1)
See e.g. https://github.com/freedomotic/freedomotic/pull/183

Any feedback is welcome.

Regards
Martin

### DIFF
--- a/mockserver-core/pom.xml
+++ b/mockserver-core/pom.xml
@@ -95,12 +95,6 @@
             <artifactId>json-schema-validator</artifactId>
         </dependency>
 
-        <!-- xml -->
-        <dependency>
-            <groupId>xerces</groupId>
-            <artifactId>xerces</artifactId>
-        </dependency>
-
         <!-- commons & guava -->
         <dependency>
             <groupId>org.apache.commons</groupId>
--- a/mockserver-core/src/main/java/org/mockserver/matchers/StringToXmlDocumentParser.java
+++ b/mockserver-core/src/main/java/org/mockserver/matchers/StringToXmlDocumentParser.java
@@ -1,9 +1,5 @@
 package org.mockserver.matchers;
 
-import com.google.common.base.Charsets;
-import org.apache.xml.serialize.Method;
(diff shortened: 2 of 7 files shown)


jamesdbloom:
This PR makes the build fail on snap-ci, please fix the issues before this can be merged

nitram509:
@jamesdbloom 
Sure, I want to fix the errors in the test, but may I ask for your help?
Because I've run the maven tests before may commit and they are also failing.
E.g. I don't get a clue, why my changes trigger a BindException "address already in use"
Any idea?

nitram509:
@jamesdbloom 
The build seems not to be stable. After a couple of unrelated commits, it finally passed.
Are there some tests running in parallel?

jamesdbloom:
as far as I'm aware the build should be stable, all the recent failures have been for known issues

jamesdbloom:
I ran the last build twice and they passed both times so it looks good to me

nitram509:
All right.
Thank you very much for your assistance.

jamesdbloom:
I'm sorry but reverting this merge because it is causing the build to fail, I think the changed version of XML parsing is interfering with the maven plugin

nitram509:
Indeed, the new XML to string converting works different than before.
Before was wrong handling of indentation (only first level was indented).
Now pretty print works correct.
This is why I needed to adopt some tests (means XML strings inside).

What test is failing?

3791:
I'll take a look at this in more over the next week or so and see if in fact it was a random build issue. It may be that since the build moved to snap-ci it has become more unstable.

PR #288, opened 2016-10-22
Update json dependency

radixhound:
In order to get the ruby client working with my Rails 5 based project, I had to remove the json dependency constraints as well as activesupport and rest_client constraints since I was locked to newer versions of these.

In theory this client should be designed to work with whatever json package is used by the project that it gets included in, rather than bringing in it's own json dependency.
- [x] _specs pass with this change_ except for one spec that seems to depend on launching an actual server to test against. 
- [x] _client works with my Rails 5 project_
- [ ] maybe would be a good idea to update specs to test against different json gems?

NB: the specs call the `to_json` method in `Enum`. My project calls the `as_json` method in `Enum` so the `as_json` case is not currently covered in the specs.

Let me know if this seems like a good direction. I couldn't tell if this was the final version of the ruby client or if there's still a new version in the works. Cheers!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ jdk:
   - oraclejdk7
 
 rvm:
-  - "2.0.0"
+  - "2.2.3"
 
 # whitelist
 branches:
--- a/mockserver-client-ruby/lib/mockserver-client.rb
+++ b/mockserver-client-ruby/lib/mockserver-client.rb
@@ -2,16 +2,3 @@
 require_relative './mockserver/version'
 require_relative './mockserver/mock_server_client'
 require_relative './mockserver/proxy_client'
-
-# Setup serialization correctly with multi_json
-require 'json/pure'
-
-# To fix serialization bugs. See: http://prettystatemachine.blogspot.com/2010/09/typeerrors-in-tojson-make-me-briefly.html
-class Fixnum
-  def to_json(_)
-    to_s
-  end
-end
-
-require 'multi_json'
-MultiJson.use(:json_pure)
--- a/mockserver-client-ruby/lib/mockserver/model/enum.rb
+++ b/mockserver-client-ruby/lib/mockserver/model/enum.rb
@@ -33,6 +33,14 @@ module MockServer::Model
     def to_s
       @value.to_s
     end
+
+    def as_json(_)
+      @value.to_s
+    end
+
+    def to_json(_)
+      "\"#{@value}\""
+    end
   end
 
   # Subclass of Enum that has a list of symbols as allowed values.
--- a/mockserver-client-ruby/mockserver-client.gemspec
+++ b/mockserver-client-ruby/mockserver-client.gemspec
@@ -28,10 +28,9 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'rubocop', '~> 0.23'
 
   spec.add_dependency 'hashie', '~> 3.0'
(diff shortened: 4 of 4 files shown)


jamesdbloom:
This PR can't be merged until the build is fixed, thanks.

radixhound:
@jamesdbloom I spent a little while with this and was unable to make headway. It seems like there is some configuration for the Snap CI build that I don't have access to? It needs to be updated to install a new ruby version on the CI runner. Things I tried:

- Update travis.yml to bump the ruby version
- Ran the tests locally (there was one failure but unrelated to the ruby client)
- Ran the tests on my own Snap CI - (this didn't run ruby tests at all since it didn't auto-detect it)

If you can give me an idea how I can work with that Snap CI build, I'm happy to keep working on it.

jamesdbloom:
The snap-ci build could be updated, however if we update the version of ruby that'll force users of the client to also upgrade.

For MockServer there is a policy of using the lowest version feasible to ensure that people working in restrictive commercial environments that can't easily upgrade their dependencies can continue to use the latest version of the libraries.

Can I please understand more about why we need this upgrade.

Thanks

radixhound:
Hmm thinking this through from the current point:

- The ruby upgrade is needed because the latest version of `active_support` needs it
- ActiveSupport needs it because removing the `~> 4.1` restriction causes it to install `5.0.0.1`
- `active_support` in the gemspec needs to be updated from `~> 4.1` because that will lock it to the 4.x range and keep it from working for Rails 5 projects.

Perhaps we could update tests to explicitly use ActiveSupport 4.x then the tests will run against that version and the ruby upgrade won't be necessary?

Or perhaps make a separate Rails 5 compatible branch?

Or maybe the client code could be updated to not depend on ActiveSupport?

Generally, I think the direction of removing the json restrictions will only help to make it more compatible. The tricky part there is how to test with the various possible json gem options out there and make sure they still work? Probably best to start by setting up the test environment to have certain gems installed and at least cover the use cases it had before this change to ensure that if people had it working in a given environment, it should continue working there?

That's a lot of question marks... Sorry for generating more questions than answers. I'm happy to tackle one or more of these solutions though once I have an idea what you'd like to see.

ovidiubite:
@jamesdbloom @radixhound Hey. guys. I read all the comments from this issue and it seems that for a good period of time nothing happened. Is this PR abandoned? I have the same issue with integrating the gem with a Rails 5 application due to the activesupport dependencies. I think it's a big deal to not be able to use it with Rails 5 and we should definitely do something to fix it.

radixhound:
@ovidiubite This branch is working very well for us on our Rails 5 project. I'd love to help get this update merged in but my hands are tied without some guidance.

ovidiubite:
@radixhound I think I'm also gonna use your branch until we manage to merge the update. Another potential issue for this repo is that end of life for Snap Ci [https://blog.snap-ci.com/blog/2017/02/06/2017-02-06-snap-announcement/]() but this is another topic.

jamesdbloom:
Unfortunately this PR can't be merged as its failing the build.

jamesdbloom:
On the topic of Snap CI no longer supporting builds, I'm looking at move back to Travis or drone.io to resolve that.

jamesdbloom:
the ruby client has moved to https://github.com/jamesdbloom/mockserver-client-ruby please raised any PRs for this client on that repository

PR #304, opened 2016-12-16
bump rubygem version

girak:
@jamesdbloom it's about time to release `1.0.9` and build gem, so we can get mockserver client working again with later versions of rails!


jamesdbloom:
the ruby client has moved to https://github.com/jamesdbloom/mockserver-client-ruby please raised any PRs for this client on that repository

PR #307, opened 2017-01-19
Add possibility to create expectations from dumped request log

masooh:
- add trigger and encoder to dump filtered request/response in single files
- extend MockServerClient for using these dumps

### DIFF
--- a/mockserver-client-java/src/main/java/org/mockserver/client/server/MockServerClient.java
+++ b/mockserver-client-java/src/main/java/org/mockserver/client/server/MockServerClient.java
@@ -1,6 +1,8 @@
 package org.mockserver.client.server;
 
 import com.google.common.base.Charsets;
+
+import org.apache.commons.io.IOUtils;
 import org.mockserver.client.AbstractClient;
 import org.mockserver.client.netty.SocketConnectionException;
 import org.mockserver.matchers.TimeToLive;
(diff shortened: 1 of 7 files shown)


masooh:
@jamesdbloom This pull request is mainly a starting point for a discussion. I would be glad to get your feedback. The aim is to speed up the process of creating prepared mockservers. My attempt:
1. Set Proxy for test with real endpoint and dump request/response
2. Remove proxy and initialize MockServer with dumped requests

jamesdbloom:
It is deliberate that expectations are not returned by the proxy or that the MockServer does not support setting expectations from a Proxy request/response dump. This is because of the anti-pattern of record and replay for recording proxies such as betamax. 

The key issue is that if test mocks are created like that the mock data becomes over complex and brittle. These types of mock are typically hard to manage and often shared across multiple tests. For example if one or more tests are created using record and replay mocks if the data mocks need to be re-recorded (i.e. due to a new field being added to a response) then the test typically has to be complete re-written.

The proxy aims to avoid the record - replay anti-pattern by providing information about what request where made so expectations can be handle crafted easily.

For this reason I am going to close this PR but I'm happy to discuss this further.  If you have a different point of view please add extra comments.

masooh:
Hmm, I see your point. I try to explain my intention.
The goal is not to use the recording as is like tapes in betamax. The recording should only *facilitate* the creation of the expectations. I described the desired workflow above too simplified. In detail:

1. Set Proxy for test with real endpoint and dump request/response. Then you have a json file your log folder for every request/response pair. `TriggerAlwaysPolicy` creates a separate file for each file. Otherwise you would have to search in one big log file.
1. Select the json files you need for your test. 
1. Remove the unnecessary parts of the expectations. Here helps the `FilterExpectationEncoder` which can remove headers and cookies.
1. If necessary modify some parts of the expectations.
1. Copy these prepared json files in your test resources folder.
1. Create the expectations with one or more of the json files (`MockServerClient.createExpectationFromJson()`)

In this detail it seems easier to just write the expectation manually, but this is more error prone and takes more time.

So there are 2 differences to the existing approach:

1. Expectations are stored in JSON. This is good if there are complex expectations or long elements which can exceed line limit.
1. In my opinion its a good thing to reuse the expectations if they are not a complete tape of severeal requests/response pairs, but only a single pair.

PR #314, opened 2017-02-16
Fixes default charset for utf file types for example application/json to UTF-8

diwakarbhatt123:
This pull request aims at solving the issue for default charset of application/json being ISO-8859-1 instead of UTF as described in this issue.
https://github.com/jamesdbloom/mockserver/issues/162

The default charset here for UTF file type is set to UTF-8 unless specified otherwise in the charset parameter.


matthewjackowski:
I'd like to point out that the current logic isn't 'wrong' per se.  But it can cause the mockserver to give a surprising response, especially when using it in proxy mode to mock a third-party API that is rather "un-specific".  
The intent with this PR is to add a more declarative level of control before defaulting to ISO.

PR #341, opened 2017-06-01
Support CORS with credentials

aarongraham:
Currently mockserver will respond with the CORS header `Access-Control-Allow-Origin: "*"` on all CORS requests. Unfortunately `*` is not allowed by the browser if credentials are going to be sent on the request. [MDN explaination](https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Requests_with_credentials).

I've had a go at adding support for this use case into mockserver. I thought the safest thing would be to add this functionality under a new config flag. 

Is this something you would consider merging in? Anything you will need to see changed first?

### DIFF
--- a/mockserver-core/src/main/java/org/mockserver/configuration/ConfigurationProperties.java
+++ b/mockserver-core/src/main/java/org/mockserver/configuration/ConfigurationProperties.java
@@ -61,6 +61,15 @@ public class ConfigurationProperties {
         System.setProperty("mockserver.enableCORSForAllResponses", "" + enableCORSForAPI);
     }
 
+    // credentials config for CORS
+    public static boolean enableCredentialsForCORS() {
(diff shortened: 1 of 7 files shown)


jamesdbloom:
This PR looks good and I'm happy to merge in principle, just a couple of points / questions:
1. why is the default for `mockserver.enableCredentialsForCORS` false?
2. do we need `mockserver.enableCredentialsForCORS` could it not just always be true, so that the Origin value is always populated dynamically instead of using `*`
3. this now has merge conflicts due to other development can you please update the PR to resolve these

jamesdbloom:
Can you please resolve the conflicts for this PR.  Then it can be merged.

aarongraham:
Sorry I'm not going to have time to work on this for quite a while. If I re-visit it I will open a new PR.

PR #351, opened 2017-06-27
MockServerClient and EchoServer being Closeable

SchulteMarkus:
MockServerClient (and EchoServer) should be a [Closeable](https://docs.oracle.com/javase/7/docs/api/java/io/Closeable.html), so it can be used as a resource in the [try-with-resource statement](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html). This improves cleanup of mockserver.

### DIFF
--- a/mockserver-client-java/pom.xml
+++ b/mockserver-client-java/pom.xml
@@ -3,7 +3,7 @@
     <parent>
         <groupId>org.mock-server</groupId>
         <artifactId>mockserver</artifactId>
-        <version>3.10.9-SNAPSHOT</version>
+        <version>4.0.0-SNAPSHOT</version>
     </parent>
 
     <modelVersion>4.0.0</modelVersion>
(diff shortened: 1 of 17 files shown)


jamesdbloom:
It is not possible to merge your change because you have change the version in the pom.xml and you have upgraded MockServer from java 1.6 to 1.7.

MockServer is deliberately implemented in Java 1.6 because several users can only support Java 1.6 in there legacy environments.

It is however a good idea for the client's to implement the Closeable interface, but to implement this does not require Java 1.7 for the test.

SchulteMarkus:
Hmm, sounds like a possible solution, but would suffer the "integration-test", which verifies MockServer is a "working" resource. But you are right, this might not worth it (having to go to Java7).
Closing here, for clean git-history, there should be a new PR.

PR #370, opened 2017-09-26
Remove obsolete Xerces dependency

alexeil:
In my project I use mockServer and log4j2. When I upgraded log4j2 to 2.9.1 I got a nasty error "java.lang.AbstractMethodError: javax.xml.parsers.DocumentBuilderFactory.setFeature(Ljava/lang/String;Z)V" after some researches, that comes from old version of Xerces as the one used in mockServer.

So I had to do the same workaround as https://github.com/jamesdbloom/mockserver/issues/362.

I propose this pull request to remove Xerces dependency. As far as I understood mockServer's code I think this dependency is no longer required.

### DIFF
--- a/mockserver-core/pom.xml
+++ b/mockserver-core/pom.xml
@@ -91,12 +91,6 @@
             <artifactId>json-schema-validator</artifactId>
         </dependency>
 
-        <!-- xml -->
-        <dependency>
-            <groupId>xerces</groupId>
-            <artifactId>xerces</artifactId>
-        </dependency>
-
         <!-- commons & guava -->
         <dependency>
             <groupId>org.apache.commons</groupId>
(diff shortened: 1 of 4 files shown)


alexeil:
@jamesdbloom can you tell me why my build is failing. 

By the way, this code is not meant to be built on Windows, i can't run the unit test properly, that's a shame.

jamesdbloom:
The build failed due to a random reason.  For some unknown reason about 1/10 build fail in the maven plugin integration tests.

jamesdbloom:
What can't you run on Windows?  There has been some effort to ensure that this runs on Windows but I haven't had time to install a Windows VM and test it properly.

jamesdbloom:
Can you please raise a bug / issue with the specific reasons why the build doesn't run on Windows.

PR #376, opened 2017-11-04
Update run_mockserver.sh to allow change the log level through docker-compose

allon2005:
…-compose by using "environment: - mockserverLogLevel=WARN(of any other log level)". If no such "- mockserverLogLevel" entry present in docker-compose, it will then use the default logLevel "INFO".


allon2005:
a sample docker-compose.yml regarding changing log level will be like this:


mockserver
  image: jamesdbloom/mockserver:latest
  ports:
  - 1080:1080
  environment:
  - mockserverLogLevel=WARN

Hope the change can be merged soon and new docker image can be generated and shared.  

thanks!

PR #395, opened 2017-12-07
Add .editorconfig support for codestyle.

arudge:
This project indents and styling differs from my default settings, so adding in an .editorconfig (http://editorconfig.org/) file.
This keeps the project specific formatting.
.editorconfig is adopted by a bunch of editors and IDEs, but I specifically use Intellij.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,6 @@
+[*]
+charset=utf-8
+end_of_line=lf
+insert_final_newline=true
+indent_style=space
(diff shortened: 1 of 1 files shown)


jamesdbloom:
The build failed due to a socket connectivity thrown by the JVM, this occasionally happens due to the massive number of sockets that are opened and closed during the build.  I'm just re-running the build to fix this then I'll merge.

PR #415, opened 2018-01-15
Fix non-convergent transitive dependencies.

brcolow:
Mockserver has some [non-convergent transitive dependencies](https://maven.apache.org/enforcer/enforcer-rules/dependencyConvergence.html).

If there are some breaking public API changes that Mockserver uses and the manually excluded dependencies don't, this will break things. This is usually somewhat rare, especially when those dependencies follow semver.

We should be able to tell from the CI infrastructure if this breaks thing as it will trigger some NoClassDefFoundErrors, or similar exceptions.

The rationale for this PR is that, in one of my projects, I use the maven-enforcer-plugin `<dependencyConvergence/>` rule, and Mockserver is the only dependency (for my project) that doesn't have convergent transitive dependencies. Outside of the personal gain from this being merged, I think it is considered good practice for third party libraries to be non-convergence free.

Hopefully this makes sense, it's a somewhat obscure thing.

### DIFF
--- a/mockserver-core/pom.xml
+++ b/mockserver-core/pom.xml
@@ -71,6 +71,12 @@
         <dependency>
             <groupId>org.apache.velocity</groupId>
             <artifactId>velocity-engine-core</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.apache.commons</groupId>
+                    <artifactId>commons-lang3</artifactId>
+                </exclusion>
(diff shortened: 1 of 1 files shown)


jamesdbloom:
Thanks I totally agree, also its helpful to exclude unused dependencies for other reasons, such as jar-with-dependencies will be smaller and quicker to build.

PR #419, opened 2018-01-18
Add PATCH to CORS allow methods

dougdragon:
This PR adds `PATCH` to the list of CORS allow methods. We're using the docker image and unless I'm mistaken, there's no way to disable CORS support for REST API.

### DIFF
--- a/mockserver-core/src/main/java/org/mockserver/cors/CORSHeaders.java
+++ b/mockserver-core/src/main/java/org/mockserver/cors/CORSHeaders.java
@@ -33,7 +33,7 @@ public class CORSHeaders {
         } else {
             setHeaderIfNotAlreadyExists(response, HttpHeaderNames.ACCESS_CONTROL_ALLOW_ORIGIN.toString(), ANY_ORIGIN);
         }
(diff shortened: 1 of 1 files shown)


dougdragon:
@jamesdbloom First PR, be polite 😄

jamesdbloom:
I agree with the change you have made but you need to run the build as some tests are failing with that change.  Alternatively I can make the change and the fix the build directly in the master branch without a PR.  I'm happy either way as you prefer.

dougdragon:
We can't see the test result, if you don't mind making the change that would be great and I'd be happy to close this PR.

dougdragon:
@jamesdbloom Thanks a ton for pushing that change into `master`! We're using the docker image and it's still pulling version `5.3.0` - how can we get an updated docker image?

dougdragon:
Hi @jamesdbloom - is there anything I can do to help with getting a new docker image with [the work you pushed to master](https://github.com/jamesdbloom/mockserver/commit/d730661f523d2f70befb6d5575f67dbc62dd8e6e) from this PR? Happy to do whatever I can to help.

jamesdbloom:
I have added a new tagged version of the MockServer build called mockserver-snapshot so please use that.  I can't due a fully release yet as there are a few more changes due to go into the next release, such as a UI that will show the in-flight requests / expectations and logs, etc.

dougdragon:
@jamesdbloom Thank you very much!

PR #423, opened 2018-01-26
Minor documentation change (flipped mockserver and proxy descriptions in 2 places)

jimkohl:
James,
sent you a note about this.  Don't know if there is a reason you didn't make sendExpectations public but tis useful to have it so.  Also changed what seems to be incorrect regarding server / proxy options (server info was w proxy; proxy info was w/ server).  Thanks!


jamesdbloom:
Good catch on the website HTML being wrong.  Can you please split that out into a separate PR because that can be merged easily.  The change to the sendException is more controversial and also has merge conflicts so I'm not sure it the right thing for now.  This is intended as an internal function and has no validation, etc.  This function is called by the `ForwardChainExpectation` as per the internal `Mockito` pattern.  I'm open to discussing / understanding why you think exposing this is a good idea and implementing a solution but I'm not sure the answer is just to expose this method.

jimkohl:
Reverted the public scope item so you can merge for now.

 I figured there was some design idea around it not being public.

Definitely need this tho!  
its a very useful function which works (I'm using it frequently from client and will continue to as I need a way to load previously stored json expectations over into the ms proxy.

jimkohl:
James I don't have a buildkite account (yet) but two checks failed.
Looks like some changes went in that I didn't have pulled, my bad ... just pushed to my fork checks are being run now.

jimkohl:
👍

jimkohl:
Just changed the title.  I'll create a separate issue for the public accessor request.

jamesdbloom:
If you can please rebase this change then I can merge it.

jimkohl:
James
Sent u a new pr!
Jim

PR #426, opened 2018-02-01
Replace Guava's Charset with JDK builtin StandardCharsets.

brcolow:
There isn't too much point to this PR except for a step in the direction of replacing all usages of Guava in Mockserver with builtin defaults. I think the ideal for a library is to be as dependency light as possible and so it is in that spirit that I offer this PR (if you are okay with the idea I shall tackle other Guava classes).

Thanks.


jamesdbloom:
I agree this makes sense.  Can you please rebase it so I can merge it.

brcolow:
@jamesdbloom Done.

PR #463, opened 2018-04-20
closes #462 - adds support for automatically comparing XMLBody using …

### LINKED ISSUE #462: Use XMLDiff to compare XMLBody
in comparing some SOAP messages, sometimes the namespaces are inverted in the serialization.  Because an ExactStringMatcher is used, this results in the expectation not being met.

Instead, use XMLDiff library for these comparisons.

danhaywood:
…XMLDiff.

### DIFF
--- a/mockserver-core/pom.xml
+++ b/mockserver-core/pom.xml
@@ -106,6 +106,10 @@
             <groupId>javax.xml.bind</groupId>
             <artifactId>jaxb-api</artifactId>
         </dependency>
+        <dependency>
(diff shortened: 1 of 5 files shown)


jamesdbloom:
I would like to give public access to the build server but that is currently not available, however here is the build log file.

[mockserver_build_484_build.log](https://github.com/jamesdbloom/mockserver/files/2114947/mockserver_build_484_build.log)

I hope that helps.

PR #474, opened 2018-05-15
Added support for jUnit5

acheaito:
Suggested change to support jUnit5 Jupiter. Created a separate module to handle this, similar to how other testing libraries addressed the issue.


acheaito:
I get 404 when clicking Details on the failed check. Can anyone advise what the reason for the failures so that I can address?

vojtoshik:
@jamesdbloom, do you have a plan to go forward with this feature? I believe it will be in high demand in nearest future as Jupiter's adoption rate growth. BTW, I had to implement similar feature today :)

jamesdbloom:
I would like to give public access to the build server but that is currently not available, however here is the build log file.

[mockserver_build_487_build.log](https://github.com/jamesdbloom/mockserver/files/2114945/mockserver_build_487_build.log)

I hope that helps.

acheaito:
@jamesdbloom Looks like the build is using JDK 7. jUnit 5 requires Java 8, so the new jupiter module has to target 8.  How do I go about having the build use JDK 8?

ketan:
Hi — is there any interest here? It's a blocker to have to use this with junit5.

I'm happy to take this PR forward if there's interest from the core-committers.

jamesdbloom:
I don’t understand why this would be a blocker to using JUnit 5.  Can you explain why?

If we switch to using Open JDK 8 then those people using 7 will no longer be able to use Mock Server.

acheaito:
@jamesdbloom - Can we make JDK 8 available for the build so that the jUnit5 module can use it while others continue to use 7 (or compile with level 7)?

jamesdbloom:
That isn't feasible because it would require splitting the build of each module into a separate build run on a different JVM.  Maven does support changing the JVM mid build because it is not feasible / possible to change the JVM running a programme mid way through the programme running.

Can you explain your exact issue fully so I can understand if there is any other alternative.

acheaito:
@jamesdbloom Sure. jUnit5 [requires Java 8 or above](https://junit.org/junit5/docs/current/user-guide/#overview-java-versions) by design. As such, my code, which creates a jUnit5 extension for mockserver, requires Java 8 or above to compile.

There is no need to switch JDKs mid build. With Maven, if the build uses JDK 8, and [the maven compiler source and target levels](https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-source-and-target.html) are set to 7, the resulting jars should be Java 7 compatible. With this approach, the parent pom can set the levels to 7 and the new jUnit module can use 8 (by overriding the compiler plugin configuration).

jamesdbloom:
I'm happy to test that, however in my previous experience although your correct about the theory I've found maven doesn't produce jars that can actually be used in Java 7 with the target and compiler level set to 7 for an 8 JVM.

flyinprogrammer:
anything we can do to help move this PR along?

damienhollis:
@jamesdbloom jupiter is gain more traction, it would be great if this pull request was able to be merged.  Do you need any assistance in getting it ready for merging?

jamesdbloom:
the tests are current failing for the PR can someone look into why they are failing?
[mockserver_build_515_build.log](https://github.com/jamesdbloom/mockserver/files/3648075/mockserver_build_515_build.log)

jamesdbloom:
Also this PR needs to be rebased before it can be merged.

jamesdbloom:
this has been merged manually so closing now

Issue #519, opened 2018-09-01
Serious performance problem in PortUnificationHandler


Vampire:
I investigated why my test cases that use the mock-server as proxy that immediately responds with 404 without forwarding need 10 seconds each, while everything should be local and fast. I even changed my test code so that only one mock-server is started when it is needed the first time and `reset` before used next and only shutdown after all test cases finished as I thought it might be the ramp-up.

What I found out is, that the problematic call is `org.mockserver.unification.PortUnificationHandler#getLocalAddresses`.
It calls `socketAddress.getCanonicalHostName()` and `socketAddress.getHostName()`.
For one of my "local" addresses, which is the "VirtualBox Host-Only Ethernet Adapter" address `192.168.99.1`, each of them needs about 5 seconds and they are called for every request, even if I do not call `reset()` inbetween.

This makes my test run need one minute with 5 tests that do a request. If I suppress these calls with a breakpoint setting, the whole tests finish in about 20 seconds instead, so it is definitely this spot where the time is lost.

While this maybe schould be made faster or at least cached so that it is only done once, is there a way as work-around to set the local address the mock-server should bind to? For running the tests it would be perfectly suitable to only bind to the loopback address. I would even prefer this, as it makes no sense to bind a service that should only be used locally to ip address 0.

Vampire:
I'm not confident enough with your code to create a PR for this one, but from a quick test, this patch seems to yield good results:

```java
diff --git a/mockserver-netty/src/main/java/org/mockserver/unification/PortUnificationHandler.java b/mockserver-netty/src/main/java/org/mockserver/unification/PortUnificationHandler.java
index a7146e3a..a7c9c33b 100644
--- a/mockserver-netty/src/main/java/org/mockserver/unification/PortUnificationHandler.java
+++ b/mockserver-netty/src/main/java/org/mockserver/unification/PortUnificationHandler.java
@@ -20,9 +20,13 @@ import org.slf4j.LoggerFactory;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Map;
 import java.util.Set;

+import static java.util.Collections.synchronizedMap;
+import static java.util.Collections.unmodifiableSet;
 import static org.mockserver.exception.ExceptionHandler.closeOnFlush;
 import static org.mockserver.exception.ExceptionHandler.shouldNotIgnoreException;
[...]

PR #522, opened 2018-09-03
Cache local address resolution in PortUnificationHandler (#519)

Vampire:


mockserver-netty/src/main/java/org/mockserver/unification/PortUnificationHandler.java
```diff
@@ -185,22 +190,32 @@ private void switchToHttp(ChannelHandlerContext ctx, ByteBuf msg) {
...
 
     private Set<String> getLocalAddresses(ChannelHandlerContext ctx) {
-        Set<String> localAddresses = new HashSet<>();
         SocketAddress localAddress = ctx.channel().localAddress();
-        if (localAddress instanceof InetSocketAddress) {
+        if (!(localAddress instanceof InetSocketAddress)) {
+            return new HashSet<>();
+        }
+
+        synchronized (localAddresses) {
```
jamesdbloom: Please replace the synchronized HashMap with a ConcurrentHashMap, which has much more efficient locked and concurrency controls. In addition once you have done this there will be no need to wrap these lines in a synchronized block which would cause performance issues when under heavy load.

jamesdbloom:
This looks good except for the couple of points I have raised, please resolve then I'll merge.  Thanks

mockserver-netty/src/main/java/org/mockserver/unification/PortUnificationHandler.java
```diff
@@ -185,22 +190,32 @@ private void switchToHttp(ChannelHandlerContext ctx, ByteBuf msg) {
     }
 
     private Set<String> getLocalAddresses(ChannelHandlerContext ctx) {
-        Set<String> localAddresses = new HashSet<>();
         SocketAddress localAddress = ctx.channel().localAddress();
-        if (localAddress instanceof InetSocketAddress) {
+        if (!(localAddress instanceof InetSocketAddress)) {
```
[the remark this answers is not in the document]
Vampire: Sure, no problem, I also replaced the other occurrence a few lines down and now have only one return statement.
I'm used to prefer preventing deep nesting by doing it that style.
But if you prefer it the other way around, that's fine with me of course. :-)

mockserver-netty/src/main/java/org/mockserver/unification/PortUnificationHandler.java
Vampire: Well, I'm not sure you correctly understand the intention.
If you want to protect iterations or query-then-insert (like done here), you have to lock the map.
With `ConcurrentHashMap` you cannot do this, as it locks on various individual buckets internally.
With `synchronizedMap` you have the explicit possibility to lock the map for iteration or query-then-insert operations.
With Java 8 I'd have used a `ConcurrentHashMap` and its `computeIfAbsent` method.
But with Java 7 this is not possible.

So to increase concurrency I could either change this to double-checked locking to only enter the `synchronized` block if actually a value has to be computed, or I can remove the `synchronized` block, then the same value gets computed multiple times when only once would have been necessary.

Waht do you prefer?

jamesdbloom: The query then insertion in this case is a safe operation to be done because at most the worst case is you update the Map twice, as you pointed out.  You are not querying modifying and then inserting so in my opinion no need to add the additional synchronisation.  Especially given the value you are adding in immutable.  

I would suggest there at worst the same value being computed multiple times is acceptable, given as you correct discovered these operations can be slow and adding synchronisation around this could cause an issue because all thread will block and the Netty thread pool is limited.  If multiple threads block it is fairly likely the Netty thread pool will not be able to satisfy incoming requests and all requests processing will slow down.  This has happened before in MockServer for similar code and so is a real concern and not just a theoretical concern.

Do you agree?

Vampire: Sure.

As I said, the alternative would be double-checked locking. (in case you don't know the term, it would mean checking whether there is a value in the map before the synchronized block and returning it if it is alread present and only entering the synchronized block if there is actually a value to be computed to prevent it from being computed multiple times)

With that in most cases the synchronized block would not be entered, as the value will already be there and thus the threads would not block, except there is really the value calculated right now.

But I agree, as it is not about having the same value on each calculation, but to save time if there is already a value calculated it is probably not tragic if the value gets calculated multiple times if two requests really need it concurrently and the parallel requests will probably have to wait about the same amount of time. And also maybe because of the more efficient implementation there is even more time saved by using the CHM, while I expect this gain to be minimal actually.

Vampire:
Update pushed

Vampire:
:-/ the changes you asked for break codeclimate

Vampire:
Ok, in case you agree on the `portExtension` `if` condition, a new version is pushed

Issue #531, opened 2018-09-13
PortUnificationHandler only works properly on fast / reliable connections


Vampire:
Currently `PortUnificationHandler#channelRead0` checks whether there are three readable bytes and then kicks off the evaluation which protocol is being used.

If no protocol can be identified the connection is closed.

But if there are less than three bytes, it just returns.

This means, for example this:

- `PU` is sent
- handler returns
- `T` is sent
- handler returns
- `PUT\n` is sent
- handler recognizes HTTP, configures the pipeline and continues

So actually the inputstream had `PUTPUT\n`, which was recognized and handled as if the first three bytes didn't exist.

So currently leading bytes may be ignored and thus requests answered incorrectly.
And additionally requests might not be recognized correctly and thus are either ignored or aborted while they would have been perfectly valid.

As far as I have understood it, the `PortUnificationHandler` must change to be statefull and collect the bytes until it can either clearly identify the protocol or clearly discard the request.

It would probably be easy to collect the bytes in a `ByteBuf` that can then be redispatched with a `fireChannelRead` after one of the `isXXX` methods recognized the protocol or after all `isXXX` methods gave up (`isSsl` e. g. needs at least `5` bytes to do its work.

But I might not be able to contribute this, as I'm not too familiar with netty and how you properly make this stateful.

jamesdbloom:
fixed in merged PR

Issue #529, opened 2018-09-12
Support for SOCKS5 authentication


Vampire:
SOCKS5 is capable of doing authentication.

Would be nice if you could also support this.

Actually currently you always answer the initial SOCKS5 requests with an auth method of `00`, meaning no auth, even if `00` was not in the list sent by the client for the server to pick from which is also not according to RFC 1928.

So there are actually two issues. 1. an auth method that was not in the list from the client should not be sent back as selected method. 2. supporting authentication would be nice, at least username / password, while the RFC demands `Compliant implementations MUST support GSSAPI and SHOULD support USERNAME/PASSWORD authentication methods.`

jamesdbloom:
fixed in merged PR

Issue #528, opened 2018-09-12
SOCKS4 / SOCKS4a not supported


Vampire:
The port unification handler recognizes SOCKS4 and SOCKS5. Unfortunately the actual handlers later on only support SOCKS5 currently. Would be nice if SOCKS4 / SOCKS4a would also be fully supported to use it for testing.

Vampire:
Actually also the port unification handler is wrong.
If version is either SOCKS4 or SOCKS5 it then tries to read the number of auth methods and then the auth methods while this is only the SOCKS5 protocol, not the SOCKS4 protocol.

jamesdbloom:
fixed in merged PR

PR #533, opened 2018-09-16
Provide SOCKS4 and SOCKS4a support and improve SOCKS5 support (#528, #529, #531)

Vampire:
- when acting as SOCKS5 proxy, provide the ability to require username / password authentication (#529)
- use a replaying decoder for the protocol detection to get as many bytes as needed (#531)
- support SOCKS4 and SOCKS4a instead of only SOCKS5 (#528)
- only select auth methods the client offered in the SOCKS5 initial response (#529)
- detect SOCKS protocols before SSL, as valid SOCKS messages can be recognized as SSL messages erroneously
- do not try to decode the amount of SOCKS5 auth schemes as auth scheme which will fail if there are more than 2 auth methods available
- do not treat SOCKS4, SOCKS4a and SOCKS5 equally when determining protocol, as their messages are quite different (#528)

Please thoroughly review this, I never did any netty stuff before and just learned what I did while doing it, so I might have made mistakes.

### DIFF
--- a/mockserver-core/src/main/java/org/mockserver/configuration/ConfigurationProperties.java
+++ b/mockserver-core/src/main/java/org/mockserver/configuration/ConfigurationProperties.java
@@ -180,8 +180,8 @@ public class ConfigurationProperties {
 
     public static void addSslSubjectAlternativeNameIps(String... newSubjectAlternativeNameIps) {
         boolean subjectAlternativeIpsModified = false;
-        for (String subjectAlternativeDomain : newSubjectAlternativeNameIps) {
-            if (ALL_SUBJECT_ALTERNATIVE_IPS.add(subjectAlternativeDomain.trim())) {
+        for (String subjectAlternativeIp : newSubjectAlternativeNameIps) {
+            if (ALL_SUBJECT_ALTERNATIVE_IPS.add(subjectAlternativeIp.trim())) {
                 subjectAlternativeIpsModified = true;
             }
         }
@@ -278,6 +278,22 @@ public class ConfigurationProperties {
 
     public static void httpProxyServerPassword(String httpProxyServerPassword) {
         System.setProperty("mockserver.httpProxyServerPassword", httpProxyServerPassword);
+    }
+
+    public static String socksProxyServerUsername() {
+        return readPropertyHierarchically("mockserver.socksProxyServerUsername", "");
+    }
+
+    public static void socksProxyServerUsername(String socksProxyServerUsername) {
+        System.setProperty("mockserver.socksProxyServerUsername", socksProxyServerUsername);
+    }
+
+    public static String socksProxyServerPassword() {
+        return readPropertyHierarchically("mockserver.socksProxyServerPassword", "");
+    }
+
+    public static void socksProxyServerPassword(String socksProxyServerPassword) {
+        System.setProperty("mockserver.socksProxyServerPassword", socksProxyServerPassword);
     }
 
     private static InetSocketAddress readInetSocketAddressProperty(String s) {
--- a/mockserver-core/src/test/java/org/mockserver/configuration/ConfigurationPropertiesTest.java
+++ b/mockserver-core/src/test/java/org/mockserver/configuration/ConfigurationPropertiesTest.java
@@ -497,4 +497,32 @@ public class ConfigurationPropertiesTest {
         assertEquals("p@ssw0rd", ConfigurationProperties.httpProxyServerPassword());
         assertEquals("p@ssw0rd", System.getProperty("mockserver.httpProxyServerPassword"));
     }
+
+    @Test
+    public void shouldSetAndReadSocksProxyServerUsername() {
+        // given
+        System.clearProperty("mockserver.socksProxyServerUsername");
+
+        // when
+        assertEquals("", ConfigurationProperties.socksProxyServerUsername());
+        ConfigurationProperties.socksProxyServerUsername("john.doe");
+
+        // then
+        assertEquals("john.doe", ConfigurationProperties.socksProxyServerUsername());
+        assertEquals("john.doe", System.getProperty("mockserver.socksProxyServerUsername"));
+    }
+
+    @Test
+    public void shouldSetAndReadSocksProxyServerPassword() {
+        // given
+        System.clearProperty("mockserver.socksProxyServerPassword");
+
+        // when
+        assertEquals("", ConfigurationProperties.socksProxyServerPassword());
+        ConfigurationProperties.socksProxyServerPassword("p@ssw0rd");
+
+        // then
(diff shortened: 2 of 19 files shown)


Vampire:
What would be nice, but not ultimately needed for my use-case would be to be able to record expectations for SOCKS mode, so that you can easily simulate error conditions, authentication and so on just like you can with HTTP requests.

But I'll leave that to you and make a new ticket for it.

Vampire:
Any idea why buildkite is refusing to build this PR?

Vampire:
@jamesdbloom unfortunately kind of not.
The build you manually started didn't report state in here, so I repushed to trigger the build, but it seems to hang again. Can you please cancel build 555?
I tried the last one that is in the log of 543 you posted and it also seems to hang here, so I can have a look into it locally.

Vampire:
@jamesdbloom the problem with the hang is a combination of problems.

The SOCKS5 implementation in Java 7 and Java 8 is broken when it gets domain type addresses back.
One of my changes was to return domain type address of length 0 where the field is to be ignored by the client anyway, as this only needs one byte opposed to 4 for the IP-type address.

The error in the Java SOCKS5 implementation causes the payload to start with a 0-byte.

The apache http client used in the test reads lines until it finds a one with a HTTP protocol.
But as the payload and thus the HTTP protocol line starts with a 0-byte, the protocol line is never found and indefinitely waited for instead which makes the test hang like that.

If I run the test with Java 9 where the implementation got fixed, the test completes properly without any problem.

So the question is what to do.
Run the tests with Java 9?
Return IP-type address and waste the few bytes to make the tests happy that depend on a buggy implementation?
Any other idea?

jamesdbloom:
For now running the test with Java 9 isn't ideal because I need to ensure they work with Java 7 and upwards.  I am planning to migrate to Java 8 soon(ish) as a major version bump but I want to fix a few more bugs before I do that otherwise users on Java 7 will be stranded.  It is possible to implement the code such that it detects the Java version and add the extra bytes if the version isn't Java 9, with a suitable comment to explain why you're doing that.  This way when we upgrade to Java 9 the code shouldn't break or just in case it does there will be a comment to indicate where to look.

The PR looks good. thanks for you work you've put in on this, except, I'm not sure why you pull our some code into MockServerUnificationInitializerDelegate.  I can't see what purpose it serves to have the code you pulled in there.  Using a delegate like that isn't a standard pattern with Netty and adds a class for little clear purpose, for example the name is not particularly meaningful because what are you delegating?

Also I happy to do a release once both your PRs are merged if you like so that you have a proper version in docker / nexus, etc.

jamesdbloom:
Also sorry for the slow responses lately I've had a lot going on in my person life recently with children, etc.  However now I'm more focused on MockServer again.

Vampire:
> It is possible to implement the code such that it detects the Java version and add the extra bytes if the version isn't Java 9, with a suitable comment to explain why you're doing that. This way when we upgrade to Java 9 the code shouldn't break or just in case it does there will be a comment to indicate where to look.

You think so?
How?
Where?
I'm not sure you got me right.
The production code is behaving fine under any version.
But the test code uses the Java proxy implementation client-wise and that has a bug that makes some tests hang or fail.
Right now I added for those 7 tests `assumeThat` calls that check the Java version and thus mark the tests as ignored if run on Java 7 or 8.

> The PR looks good. thanks for you work you've put in on this,

you're welcome

> except, I'm not sure why you pull our some code into MockServerUnificationInitializerDelegate. I can't see what purpose it serves to have the code you pulled in there. Using a delegate like that isn't a standard pattern with Netty and adds a class for little clear purpose, for example the name is not particularly meaningful because what are you delegating?

As I mentioned in one of the tickets, I didn't do any netty stuff before.
So if I should do something differently, please tell me.
The reason was (is):
- `MockServerUnificationInitializer` must be `@ChannelHandler.Sharable` as it is set as `childHandler` for the `ServerBootstrap`
- `PortUnificationHandler` is no `SimpleChannelInboundHandler<ByteBuf>` anymore, but a `ReplayingDecoder<Void>`
- becaues of that `PortUnificationHandler` cannot be `@ChannelHandler.Sharable` anymore
- because of that `MockServerUnificationInitializer` cannot be a subclass of `PortUnificationHandler` anymore
- so I made the `MockServerUnificationInitializer` delegate its actual work to the `MockServerUnificationInitializerDelegate` by replacing itself with a fresh delegate if it got added to a pipeline

> Also I happy to do a release once both your PRs are merged if you like so that you have a proper version in docker / nexus, etc.

Would be great, yeah, thanks :-)

jamesdbloom:
I do fundamentally like what you've done so I'm not trying to be critical but I do want to understand where you are coming from and discussion different design options.

I'm interested in why you changed from `SimpleChannelInboundHandler<ByteBuf>` to a `ReplayingDecoder<Void>`.  I don't believe that is necessary or correct because for example your not using the out field of the Decoder.  A Decoder is for transforming the input type from one type to another for example from `ByteBuf` to `WebSocketFrame`.  The `PortUnificationHandler` does actually do any decoding instead it sets up the pipeline.

I'm also interested in why you removed the guards for the number of bytes.  If I remember correct they are there to avoid errors when an insufficient amount of data is sent.  I believe that the newer version of the HTTP detect may have issues for some requests with small payloads.  However, to be honest I can't remember the scenarios under which that was an issue.

Vampire:
> I do fundamentally like what you've done so I'm not trying to be critical but I do want to understand where you are coming from and discussion different design options.

Sure, no problem.
I'm always open for constructive criticism, especially as I'm not too familiar with the "netty-way".

> I'm interested in why you changed from SimpleChannelInboundHandler<ByteBuf> to a ReplayingDecoder<Void>.
> ...
> I'm also interested in why you removed the guards for the number of bytes.

Both are connected.
Actually the second change was because it is not necessary anymore with the first change.
For example to detect SOCKS4 you need much more bytes than to detect a possible SOCKS5 initial request or an HTTP request. But actually the initial SOCKS4 request can have fewer or more bytes, depending on whether it is SOCKS4a or not.
And if you have a slow connection, let's just assume the worst-case scenario and the bytes come in individually with some delay inbetween.
The old detection mechanism will just assume it cannot recognize the request and deny it, as the `SimpleChannelInboundHandler` can only handle the bytes that came in already and cannot accumulate them.
The `ReplayingDecoder` on the other hand, if you request the next 10 bytes and only 5 bytes are available, it throws an exception that is known, caught, waited for more bytes to be arrived and then the pipeline is restarted with the more bytes available.
So with the `ReplayingDecoder` you don't need the byte-amount gates and the recognition will always have enough bytes to determine, as it simply waits for more bytes being available if necessary.

The only problem I see if there are fewer bytes available than requested, e. g. if the first 5 bytes could be the start of a SOCKS4 request but there are no more bytes being sent.
But in such a case I'd expect reply-timeouts to kick in and abort the connection as obviously it is nothing the mock-server can recognize.

PR #555, opened 2018-11-06
add a timestamp to the request log

samba09:
Dear Proxy Users,

I need to have the timestamps in the log of the requests.

example:

curl -X PUT "localhost:1090/retrieve?type=REQUESTS&format=JSON"

result is:

[ {
  "method" : "GET",
  "path" : "/",
  "headers" : {
    "User-Agent" : [ "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/64.0.3282.140 Safari/537.36 Edge/17.17134" ],
    "Accept-Language" : [ "en-DE" ],
    "Accept" : [ "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8" ],
    "Upgrade-Insecure-Requests" : [ "1" ],
    "Accept-Encoding" : [ "gzip, deflate" ],
    "Host" : [ "www.google.de" ],
    "Proxy-Connection" : [ "Keep-Alive" ],
    "Cookie" : [ "OGPC=19008559-4:; 1P_JAR=2018-11-06-12; CONSENT=YES+DE.en+V9; NID=142=orpBblw533hKRfECuYCJ8AG-X-ckwspfToZzr4ypuJOWOnBDIVxz2lmRyZyk109rcjD8SlkVXC7rFLs_A8Rq6mHu0xLmecn4ZR_xzDSiGCZrsML5N-KbXqZIhvdCwIuEGcz7GvUnXONbaQroMRQHbW7ux2uY3R6LTHwMScTRsaU" ],
    "content-length" : [ "0" ]
  },
  "cookies" : {
    "1P_JAR" : "2018-11-06-12",
    "CONSENT" : "YES+DE.en+V9",
    "NID" : "142=orpBblw533hKRfECuYCJ8AG-X-ckwspfToZzr4ypuJOWOnBDIVxz2lmRyZyk109rcjD8SlkVXC7rFLs_A8Rq6mHu0xLmecn4ZR_xzDSiGCZrsML5N-KbXqZIhvdCwIuEGcz7GvUnXONbaQroMRQHbW7ux2uY3R6LTHwMScTRsaU",
    "OGPC" : "19008559-4:"
  },
  "keepAlive" : true,
  "secure" : false,
  "timestamp" : 1541505971
} ]

PR #556, opened 2018-11-08
Requestlogs contain now new request like it was changed with template

samba09:


jamesdbloom:
Please note as stated in #555 the request log now has a timestamp.  Adding a timestamp to the HttpRequestDTO is not the correct place to add this because it has a very wide impact for logic that is only needed in a specific location.

In addition the RequestResponseLogEnter represents what was sent to MockServer and not what was forwarded from the proxy.  Changing this to include the templatedRequest breaks the semantics of this log entry and breaks its intended usage.

I have added the timestamp already but if you could explain your use case around the templatedRequest then we can discuss the correct place for this logic.

PR #566, opened 2018-11-27
Don't run Docker container as root

GalaxyGorilla:
Some people don't like having their containers running as root user, especially in context of k8s deployments. This patch will create and use the user 'mockserver' with UID 1000 instead.


jamesdbloom:
Yes this make perfect sense and I agree it is a good idea.

PR #595, opened 2019-02-26
Corrected Host header - no port necessary

TolikT:
Because of incorrect Host header returned response status is 404.


jamesdbloom:
But the port is required if a non-standard port is used i.e. not 443 for HTTPS or 80 for HTTP.

Do you think otherwise?

TolikT:
You're right. According to the spec it can be there. So this issue should be resolved somehow else.

jamesdbloom:
What issue are you having?

TolikT:
When I try to use client to access mockserver that is available by default 443 port, response code 404 is returned because this port is added to the Host header. Tried to remove it and got 200, everything works.

PR #613, opened 2019-04-06
Added support for LetsEncrypt to Helm Chart

mdebarros:
Suggested change to Helm charts to support LetsEncrypt. The change allows for both a custom TLS configuration and/or LetsEncrypt to be enabled.

Also moved a few things around to match semantics.

The prerequisite that you install a cert-manager:

```bash
helm install stable/cert-manager \
  --name cert-manager \
  --namespace kube-system \
  --version v0.5.2
```

P.s. Thanks for Mockserver. I have been using it for a while now and it is really useful!


.gitignore
```diff
@@ -93,3 +93,7 @@ mockserver-proxy-war/tomcat/
 mockserver-proxy-war/tomcat_proxy/
 mockserver-war/tomcat/
 mockserver-war/tomcat_mockserver/
+
+# ignored files #
+##############################
+*IGNORE*
```
mdebarros: I often add this to my repos so I can add environment config files that should be ignored by Git. I am happy to remove this if you prefer.

helm/mockserver/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 name: mockserver
-version: "5.5.1"
+version: "1.0"
```
mdebarros: I believe this change is semantically more correct and is aligned to how you attended the Helm chart to be used. Specifically that you set the container version based on the "AppVersion" which is independent of the Helm chart "Version".

jamesdbloom: The version of all artefacts is linked to reduce complexity this means that the version of the helm chart should be the same is the underlying version of MockServer.  This is important to reduce complexity both for those implementing MockServer and for users of MockServer so they can easily understand what artefact versions work with other artefact versions, i.e client, docker, helm, server, etc.

jamesdbloom: Also this value drives the generated tar file name and path

mdebarros:
@jamesdbloom Please see the update. Sorry for the delay, was on a multi-day hike with no internets \:D/

jamesdbloom:
if you can please rebase this change then I'll happily merge it

jamesdbloom:
I like the changes you have proposed here but according to Jetsstack cert-manager is only on the second alpha release and the API is not yet stable.  So I think it is too early to integrate this solution because I'm concerned about ensuring it continues to work reliably.  I am going to close this PR for now, but add an item to the backlog to integration LetsEncrypt.

PR #619, opened 2019-04-23
Fix memory leak in DashboardWebSocketServerHandler (#618)

### LINKED ISSUE #618: Memory leak in DashboardWebSocketServerHandler
<h3>Description</h3>
<p>
Mockserver goes out of memory when client sends mutliple non-keepalive requests, i.e. opening and closing connections even if reset is invoked.
</p>
<h4>Technical details</h4>
<p>
The PortUnificationHandler adds an instance of DashboardWebSocketServerHandler for client channels. The handler registers itself as a listener to mockServerMatcher and mockServerLog which are both shared across channels, but does not remove itself from their subscriptions. 
It causes multiple instances of DashboardWebSocketServerHandler being stored in their listener lists even when appropriate channels already closed.
</p>
<h3>Mockserver version</h3>
<p>5.5.1</p>

ikurovsky:
Motivation:
Instances of DashboardWebSocketServerHandler are not removed from listeners list even if appropriate channels are closed which causes a memory leak.

Modification:
Unsubscribe the handler on channel close.

Fixes #618

### DIFF
--- a/mockserver-core/src/main/java/org/mockserver/ui/MockServerEventLogNotifier.java
+++ b/mockserver-core/src/main/java/org/mockserver/ui/MockServerEventLogNotifier.java
@@ -35,4 +35,8 @@ public class MockServerEventLogNotifier extends ObjectWithReflectiveEqualsHashCo
     public void registerListener(MockServerLogListener listener) {
         listeners.add(listener);
     }
+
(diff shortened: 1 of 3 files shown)


ikurovsky:
@jamesdbloom Could you please fix the codeclimate checks? It doesn't seem to show a real problem.

Issue #661, opened 2019-08-14
CA certificate not up to date


sileht:
Hi,

I have recently updated some system to debian buster that now uses openssl 1.1.1. and the mockserver ssl certificat is no more valid.

rsa key must be 2048bit long.

While the mockserver CA SSL certificat looks good. The certificate built on the fly for custom domain looks have only 1024bit. 

```
$ curl --cacert mockserver-ca.pem https://www.example.com
curl: (60) SSL certificate problem: EE certificate key too weak
More details here: https://curl.haxx.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.

$ openssl s_client -CAfile mockserver-ca.pem -connect www.example.com:443
CONNECTED(00000003)
depth=0 CN = www.example.com, O = MockServer, L = London, ST = England, C = UK
verify error:num=66:EE certificate key too weak
verify return:1
depth=1 CN = www.mockserver.com, O = MockServer, L = London, ST = England, C = UK
verify return:1
depth=0 CN = www.example.com, O = MockServer, L = London, ST = England, C = UK
verify return:1

...

---
SSL handshake has read 2105 bytes and written 436 bytes
Verification error: EE certificate key too weak
---
New, TLSv1.2, Cipher is ECDHE-RSA-AES128-GCM-SHA256
Server public key is 1024 bit
```

Regards,

ondro2208:
Files https://github.com/mock-server/mockserver/tree/master/mockserver-core/src/main/resources/org/mockserver/socket is not up to date. I am getting ERR_CERT_WEAK_KEY

jamesdbloom:
I will update these files prior to doing the next release otherwise I'll break the current released version of MockServer.  The new release will include new logic to generate certificates that don't rely on the good but massively over-bloated bouncy castle.  I've added a note to this backlog item to update the certs https://trello.com/c/tsJDyFuy/80-replace-bouncy-castle-with-java-8-jdk-approach-for-dynamic-certs

jamesdbloom:
I'm closing the issue as I can't reproduce on the current codebase.  In addition in the new release that will be completed and release as soon as the final TLS documentation is updated, there have been extensive updates to the TLS logic.  The following features have been added:
- dynamically create local CA X509 (instead of using fixed CA in git repo) - only enabled via configuration
[...]

PR #662, opened 2019-08-14
ssl: generates 2048 RSA key

sileht:
Openssl 1.1.1 does not considere 1024bit RSA key as secure.

So this patch changes the generated key size to 2048

Closes: #661


sileht:
Anything I can to do to get this merge ?

sileht:
Thx!

ondro2208:
Files here https://github.com/mock-server/mockserver/tree/master/mockserver-core/src/main/resources/org/mockserver/socket are not up to date

PR #693, opened 2019-11-25
HttpRequestEnhancer for MockServerClient: This feature is useful, whe…

micst96:
In our project we require additional authorization header to authenticate request that goes to mockserver through our proxy. This feature allows modify request before it will be sent to mockserver so we can easily add new headers.


jamesdbloom:
I agree this makes sense, but the majority of this logic already exists in the format of the update method on the HttpRequest.  Perhaps you could tweak this so that an HttpRequest object set that contained anything that needed to updated instead of a class implementing the enhancer interface.  That would only be a minor tweak logically from what you have but would be more inline with existing code and patterns.  Do you know if there is a reason why that wouldn't meet your requirements?

micst96:
The problem with org.mockserver.model.HttpRequest#update method is that it modify existing HttpRequest object so in scenario where we want to add new header instead of replace it, the method throws NullPointerException. 
Scenario:
1. call org.mockserver.client.MockServerClient#reset method (`request().withMethod("PUT").withPath(calculatePath("reset"))` creates new object with null headers):
`public MockServerClient reset() {
        getMockServerEventBus().publish(EventType.RESET);
        sendRequest(request().withMethod("PUT").withPath(calculatePath("reset")));
        return clientClass.cast(this);
    }`

2. org.mockserver.client.MockServerClient#sendRequest updates request
```
private HttpResponse sendRequest(HttpRequest request) {
        request.update(preparedRequest);
        ...
}
```
3. We cannot use org.mockserver.model.HttpRequest#update method, because headers from request in org.mockserver.client.MockServerClient#reset method is null (`getHeaders()` returns null value) so during org.mockserver.model.HttpRequest#getHeaders throws NullPointerException.

`public HttpRequest update(HttpRequest replaceRequest) {
        ...
        for (Header header : replaceRequest.getHeaderList()) {
            getHeaders().replaceEntry(header);
        }
        ...
        }`

In my opinion solution with interface is more flexible because we can dynamically add different properties to HttpRequest, but we can add solution with protected method that will also meet our requirements and requires only two small changes:
1. add new protected method in MockServerClient 
2. use it in "sendRequest" method

In org.mockserver.client.MockServerClient#sendRequest: 
```
private HttpResponse sendRequest(HttpRequest request) {
        modifyRequest(request);
        ...
}
```
In MockServerClient:
```
    protected HttpRequest modifyRequest(HttpRequest request){
        return request;
    }
```

jamesdbloom:
Ok that sounds like a bug which I'll fix, due to recent performance improvements

jamesdbloom:
I've pushed a fix for those null pointers to master

micst96:
Could you also consider this protected method? Even with fix for update method we need somehow modify request in org.mockserver.client.MockServerClient#sendRequest:

private HttpResponse sendRequest(HttpRequest request) {
        modifyRequest(request);
        ...
}

because this method is used in every MockServerClient method when we want to send request to MockServer and it is private. We just want to modify request in this method.

jamesdbloom:
Yes, I agree that makes sense.  

My thought was at this point https://github.com/mock-server/mockserver/blob/master/mockserver-client-java/src/main/java/org/mockserver/client/MockServerClient.java#L172 that in that method we should check if a field called requestEnhancer (of type HttpRequest) had been set (i.e. was not null).  If this field had been set it could be used to update/enhance the request being sent immediately prior to sending the request.

micst96:
Ok I have changed it according to your suggestion, but there is a still problem with update() method (you can check this test to see the problem org.mockserver.client.MockServerClientTest#shouldEnhanceRequestWithAuthorizationHeader)
We can fix it changing object on which we call update(...) method to this:
`request = defaultRequestProperties.update(request);` but I still think it should also work on `request` object.

Do you prefer this solution with update() method, or the one with protected method (because as I mentioned, with protected method we need only one method with default behaviour that do nothing with request and then we can easily override it in MockServerClient subclass)? And what are other requirements that I should meet to make this change possible?

jamesdbloom:
I think in general subclassing is a much more complex approach for most people to understand so I have merged the solution you have above.  I will slightly rename the field name though as it isn't really a default it is more an override because values in the HttpRequest replace/update any existing values as per the `update` function.

jamesdbloom:
This is now in the latest SNAPSHOT version and will be released in the next RELEASE, likely in a few weeks

micst96:
Great, thank you. Additionally you can also remove org.mockserver.client.HttpRequestEnhancer interface that I forgot about.

Issue #698, opened 2019-11-28
Charset not parsed from valid Content-Type header


DDebray:
Currently the content type is parsed by the class ContentTypeMapper incorrectly. It only is able to parse the charset when its not followed by additional parameters. This is allowed according to the RFC.

https://tools.ietf.org/html/rfc7231#section-3.1.1.1
>media-type = type "/" subtype *( OWS ";" OWS parameter )
>     type       = token
>     subtype    = token
>
>   The type/subtype MAY be followed by parameters in the form of
>   name=value pairs.
>
>  parameter      = token "=" ( token / quoted-string )

One example would be 

`application/soap+xml;charset=UTF-8;action="somerandomstuff"`

jamesdbloom:
I have now fixed this, which is available in the latest SNAPSHOT version.  I will do a new release in the next few weeks which will contain this fix.

DDebray:
well done, this is far superior to my PR :). I left a comment on the commit regarding the whitespace handling.

PR #700, opened 2019-11-28
#698 fix parsing of charsets from Content-Type header

DDebray:
We ignore the additional parameters now an only take the charset value.


jamesdbloom:
Thank you for submitting the PR, however, I fixed this in a more complete way mainly by switching to the built-in Netty implementation.  I also massively improved the way MediaTypes are parsed generally because the Google Guava library was unstable and didn't support parameters.  Thank you for taking the time to submit a PR, in future you PR will likely be merged it is just this time the problem you highlighted had wider implications.

PR #732, opened 2020-02-20
fixed request body regex example

jhopperzw:
the example for regex body for Rest API is not correct. It should be "regex" not "body"


dalcantara:
This should be updated in the documentation as well: [https://www.mock-server.com/mock_server/creating_expectations.html#button_match_request_by_body_with_xpath](https://www.mock-server.com/mock_server/creating_expectations.html#button_match_request_by_body_with_xpath)

jamesdbloom:
done

PR #756, opened 2020-04-06
Add apiVersion to Chart.yaml

jakubjab:
Make helm lint pass

### DIFF
--- a/helm/mockserver/Chart.yaml
+++ b/helm/mockserver/Chart.yaml
@@ -1,3 +1,4 @@
+apiVersion: v1
 name: mockserver
 version: "5.10.0"
 appVersion: "5.10.0"


jamesdbloom:
Can you please change the version value to “5.10.0” then I’ll merge the PR

jakubjab:
Did you mean change from “5.10.0” to “5.10.1”? Should I change only the chart version and keep the app version the same, or you prefer to keep both in sync?

jamesdbloom:
I would like to keep everything in sync as the API always changes with new releases of the app and the chart contains the app version. In general for API driven software that is almost always the case (although not 100% of the time).

Please change the version to 5.10.0 because the release process relies on the chart having the provision released version not the SNAPSHOT version.

jakubjab:
I'm sorry, I don't understand, I'm new to helm charts. The appVersion and version are already 5.10.0. I'm just adding apiVersion, which is required by helm and can be either v1 or v2. It's not a breaking change and I'm not even sure if it can be considered as a new chart version, as the chart itself remains the same.

jamesdbloom:
Yes sorry I was being stupid I was thinking it was the external API version not the helm API version, I think that is a new field since helm 3 was released?

What is v1 and v2 is that for helm version 2 and 3 respectively?

jakubjab:
Yes, v2 means that a chart requires Helm 3, which I think is not the case here.
See https://helm.sh/docs/topics/charts/#the-apiversion-field.

PR #763, opened 2020-04-14
Adding support for global server state between requests, session.

SecuriFresh:
Session is implemented as a Map which is persisted in memory.
Session entries can be set by a http response of a matched expectation.
Existing session entries can be used as an expecation matcher.
This commit includes:
1. adding Session object to HttpRequest and HttpResponse classes.
2. adding matcher for comparing the server session entries against the session entries in an expecation request
3. session entries in matched expecation response are merged to server session - explicit empty session in matched response clears the entire session state.


SecuriFresh:
still didn't add unit tests for this feature

jamesdbloom:
Closing as no activity for two year and no unit tests and not rebased.

Issue #762, opened 2020-04-14
HttpClassCallback only allows a return value of HttpRequest.


andrewfinnell:
**Describe the issue**
There are example of using a Class callback for the Response of an Expectation. However, because the class HttpClassCallback uses `HttpRequest` for it's Generic Parameter, it only allows for the callback to return an `HttpRequest`, thus it's impossible to have a callback for a Response.

**What you are trying to do**
Return a custom Response similar to this example: https://www.mock-server.com/mock_server/creating_expectations.html#button_response_class_callback

I could not get this example to compile.

**MockServer version**
5.10.0

**To Reproduce**

```
public class MyResponseCallback implements ExpectationResponseCallback {
    @Override
    public HttpResponse handle(HttpRequest httpRequest) throws Exception {
        return response()
                        .withStatusCode(status)
                        .withContentType(MediaType.APPLICATION_JSON)
                        .withBody("{ }");
    }
}

ClientAndServer clientAndServer = ClientAndServer.startClientAndServer(10000);
clientAndServer.
        when(request()
                .withSecure(tlsEnabled)
                .withMethod("GET")
                .withPath("/test"), Times.unlimited()).
        respond(
                callback()
                    .withCallbackClass(MyResponseCallback.class)); 
```

The line `.withCallbackClass(MyResponseCallback.class)` will never work as it expects a method signature of `public HttpRequest handle(HttpRequest httpRequest)`.

**Expected behaviour**
I would suspect that HttpClassCallback should be as follows:

```
 public static HttpClassCallback callback(Class<? extends ExpectationCallback<HttpObject>> callbackClass) {
        return new HttpClassCallback().withCallbackClass(callbackClass);
    }
```

**MockServer Log**
Log output, as INFO level (or lower)

**Workaround**

It appears the ExpectationResponseCallback can be created directly as a parameter to the `response()` call. This provides everything I need, so this ticket is not urgent. I just thought you'd want to know about the documentation versus implementation.

pilhuhn:
@andrewfinnell Can you describe your workaround in more detail please?

jamesdbloom:
There is a work around for this currently, as all the integration tests pass the class as a string which works.

However I have now fixed this issue which will be in the next release which will be released in the next couple of days.

PR #764, opened 2020-04-14
Update HttpClassCallback to support HttpResponse return.

andrewfinnell:
- This addresses the Issue https://github.com/mock-server/mockserver/issues/762


jamesdbloom:
This has already been fixed a different way.

PR #793, opened 2020-06-03
Enable multiple instances spawned in the same directory by adding suffix to keystore

bkabrda:
We have a pretty large testsuite that we run from parallel forks via surefire plugin. When recording, we spawn an instance of mockserver per fork. This results in various issues that all come down to both instances trying to read/write/create `mockserver_keystorejks` file. This PR solves this by allowing to add a custom suffix to the keystore file, thus avoiding the conflicts.

(There are obviously multiple ways to do this, including generating a unique temporary file for each new KeyStore, so I'm willing to work on this further if you think it'd make sense)

### DIFF
--- a/mockserver-core/src/main/java/org/mockserver/socket/tls/KeyStoreFactory.java
+++ b/mockserver-core/src/main/java/org/mockserver/socket/tls/KeyStoreFactory.java
@@ -28,6 +28,7 @@ public class KeyStoreFactory {
     public static final String KEY_STORE_CERT_ALIAS = "mockserver-client-cert";
     public static final String KEY_STORE_CA_ALIAS = "mockserver-ca-cert";
     public static final String KEY_STORE_FILE_NAME = "mockserver_keystore" + KEY_STORE_TYPE;
+    private final String keyStoreFilenameSuffix;
     /**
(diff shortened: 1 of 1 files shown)


jamesdbloom:
Which version of MockServer are you using because the KeyStoreFactory is only provided as a convenience class and it is not used internally by MockServer?

bkabrda:
@jamesdbloom I reproduced this on 5.9, 5.10 and latest master as of yesterday, but I think the change you just did in https://github.com/mock-server/mockserver/commit/34e792f47c23bf0e96e4a29ee331aabdaf7bce02 will probably fix that, I'll try it out.

The failure that we're seeing is not in mockserver, it's in the test code that makes client use KeyStoreFactory to be able to access the mockserver.

jamesdbloom:
Yes that makes sense, hopefully that code change I made should resolve your issue.  I believe that change is already in the SNAPSHOT and will be released in a couple of weeks after some other changes have been made.

bkabrda:
Yeah, I managed to verify that your change fixed my issue. Thanks a lot and I'm looking forward to the new release :).

Closing this PR now.

PR #803, opened 2020-06-20
added --create-namespace to helm

survivant:
with --create-namespace it will create the namespace.  Available since Helm 3.2.x


jamesdbloom:
Although I agree with the use of this flag helm 3.2.x has only been available since April 22nd so I think it is a little too early to add this flag to the documentation as for anyone not using >3.2.x they will get the following error
```bash
Error: unknown flag: --create-namespace
```

I'm happy to leave this PR here and merge it in 3/4 months time once that version is likely more widely used.

jamesdbloom:
please rebase so this can be merged

survivant:
do I have to do something ?

PR #868, opened 2020-09-17
Remove duplicated code in NottableString

acanda:
The removed lines are the same as lines 102-105.


acanda:
Never mind. Lines 102-105 and 110-113 are identical but both are necessary. I'm closing the PR.

PR #947, opened 2020-12-12
Fix flaky tests 

JoshSu:
I detected a few flaky tests using nondex framework, and I fixed those flaky tests.

Please see the description of the framework in here, and please let me know if there is any question

https://github.com/TestingResearchIllinois/NonDex

### DIFF
--- a/mockserver-core/pom.xml
+++ b/mockserver-core/pom.xml
@@ -15,6 +15,13 @@
     <url>http://www.mock-server.com</url>
 
     <dependencies>
+
+        <dependency>
+            <groupId>org.skyscreamer</groupId>
+            <artifactId>jsonassert</artifactId>
+            <version>1.5.0</version>
+            <scope>test</scope>
+        </dependency>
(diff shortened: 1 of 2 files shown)


jamesdbloom:

jamesdbloom:
I've not seen these test fail due to the order of the fields changes across multiple machines (which are all Mac or Linux) but I do accept that it is very possible for that to have happened. So I can understand better do you know under what circumstances the field order changed?

I have now added the following line objectMapper.configure(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY, true) to configure the creation of all ObjectMapper objects so I don't believe any tests should be flaky due to json field order. Can you please recheck if you find the test flaky so I can close this PR as believe it is no longer needed.

jamesdbloom:
I'm going to close this PR because I assume the use of `objectMapper.configure(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY, true)` has resolved any field ordering issues

Issue #505, opened 2018-07-18
CORS: Allow Credentials


antoninvogt:
I am having trouble testing an application which makes CORS requests with credentials.

I know this has been discussed in #401, but it appears that the solution proposed there does not work everywhere.

This is what I believe happens in our case:

1. The application runs on `localhost:1234`.
1. The app makes a cross-origin request to `localhost:2345` with a [credentials mode](https://fetch.spec.whatwg.org/#concept-request-credentials-mode) of `include`, using `fetch("http://localhost:2345/some/resource", { credentials: "include" })`.
1. Mock-Server does not recognize that the request includes credentials and returns a response with its global default CORS headers.
    * The response has an `Access-Control-Allow-Origin` header value of `*`.
    * The response has no `Access-Control-Allow-Credentials` header.
1. The browser determines that it is not safe to give web content access to the response.

The browser turns down the response with the following error message.

```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at ‘http://localhost:2345/some/resource’. (Reason: Credential is not supported if the CORS header ‘Access-Control-Allow-Origin’ is ‘*’).
```

(See MDN's [CORS - Credentialed requests and wildcards](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Credentialed_requests_and_wildcards) for details.)

Regardless, the response would have been hidden from web content because the `Access-Control-Allow-Credentials` header is missing (see [CORS - Access-Control-Allow-Credentials](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Access-Control-Allow-Credentials)).

## Why Does the Existing Solution Not Work? ##

If I understand the change in https://github.com/jamesdbloom/mockserver/commit/16647d8f7810f088a1060c8b0dedf276115337de#diff-6649b9bf089ed61ac03004199e753e0bR30 correctly, the idea was to mirror the specific origin and append the credentials header if the request comes with an `Access-Control-Allow-Credentials` header.

However, this header is not actually sent by the client, but by the server to tell the client that the response is safe to use (see [Fetch spec - 3.2.5. CORS protocol and credentials](https://fetch.spec.whatwg.org/#cors-protocol-and-credentials)).

This is what the actual request looks like, triggered by the browser's native `fetch` API:

```
GET /some/resource HTTP/1.1
Host: localhost:2345
Connection: keep-alive
Origin: http://localhost:1234
User-Agent: <SOME-USER-AGENT>
[...]

Issue #401, opened 2018-01-04
CORS issue when doing XMLHttpRequest having withCredentials


martinhristov:
"Response to preflight request doesn't pass access control check: The value of the 'Access-Control-Allow-Origin' header in the response must not be the wildcard '*' when the request's credentials mode is 'include'. Origin 'http://localhost:5555' is therefore not allowed access. The credentials mode of requests initiated by the XMLHttpRequest is controlled by the withCredentials attribute."

A possible solution to this issue is to dynamically set the value of Access-Control-Allow-Origin in the response headers to the one of Origin request header.

jamesdbloom:
Good point, I can fix this.

jamesdbloom:
On further research I'm not sure what you are saying is correct, because the withCredentials attribute is a client side attribute and is not controlled on the server side.  Can you please provide a link the CORS specification describing this behaviour as I can't find any mention of it in the CORS RFC.  https://www.w3.org/TR/cors/#access-control-allow-credentials-response-header

jamesdbloom:
I have found details of how this should work and have implemented the support for `access-control-allow-credentials` and `null` `origin` headers.

cristopher0820:
Hi bro, i couldnt solve this problem. I have a client code and i cant take the cookies (credentials) from a external server by using web services. 
Im using javascript.
Can you help me pls.

Best regards.

jamesdbloom:
Have you tried using the CORS configuration settings in MockServer see: http://mock-server.com/mock_server/configuration_properties.html#button_configuration_enable_cors_for_all_responses

eweap:
@jamesdbloom Hi, I think this issue should be re-opened, there is still no configuration option to tell mock-server to stop using `access-control-allow-origin: *` 

Something like `mockserver.corsAllowOrigin` would be great !

For the moment I still don't found any workaround to this error:
```
Access to XMLHttpRequest at 'https://api.mysite.home:1080/token' from origin 'https://app.mysite.home:8000' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: The value of the 'Access-Control-Allow-Origin' header in the response must not be the wildcard '*' when the request's credentials mode is 'include'. The credentials mode of requests initiated by the XMLHttpRequest is controlled by the withCredentials attribute.
```

Best regards

PR #971, opened 2021-01-19
Fix CORS with credentials

cecedille1:
Mockserver expects a `Access-Control-Allow-Credentials in their preflight request.` to include the `Access-Control-Allow-Credentials: true | false` header. But the header is never sent. Additionaly when `Access-Control-Allow-Credentials`, the header `
access-control-allow-origin` cannot be `'*'`

cf #401 #505


jamesdbloom:

cecedille1:
Thanks James for maintaining this software :1st_place_medal:

PR #979, opened 2021-02-04
Log the ECMA262 warning instead of using stderr

candrews:


candrews:
The build errors appear to be unrelated to this change.

jamesdbloom:
stderr is used here because the library that logs a warning about this beyond my control also uses stderr.  I can see why you change could make sense but can you explain why you feel it is a good idea.

Issue #988, opened 2021-02-22
HttpClassCallback static builder has wrong signature


reitzig:
**Describe the issue**

`HttpClassCallback#callback(java.lang.Class<? extends ExpectationCallback<HttpRequest>>)` has wrong signature.
The parameter should be of type `Class<? extends ExpectationCallback<HttpResponse>`.

**What you are trying to do**

I'm trying to program an expectation initializer. Porting code that used `MockClientServer`, specifically `ForwardChainExpectation#respond(ExpectationResponseCallback)`, I was expecting to use the same class implementing `ExpectationResponseCallback`. The JavaDoc suggests that should work; at least, `Expectation#thenRespond(HttpClassCallback)` leads me to:

![image](https://user-images.githubusercontent.com/4246780/108741938-7e724080-7537-11eb-9de0-654d2ebebadd.png)

However, `ExpectationResponseCallback` implements `ExpectationCallback<HttpResponse>` which clashes with `ExpectationCallback<HttpRequest>` in the signature.

**MockServer version**
5.11.2

reitzig:
:clap: @jamesdbloom Thanks!

PR #989, opened 2021-02-22
Fix issue #988 by changing callback(Class<? extends ExpectationCallback<HttpRequest>> callbackClass) signature

reitzig:


reitzig:
Ah, this seems to break code that relies on the "wrong" (?) signature. Too bad!

jamesdbloom:
Closing because this change breaks code that relies on the current signature

PR #1009, opened 2021-04-01
Helm: omit clusterIP from the Service entirely if not provided

rmmorrison:
When clusterIP is provided as an empty string to Helm's three-way patch mechanism during an upgrade, Helm sees the Kubernetes auto-assigned value in the cluster but thinks the desired value is "" (empty string), as that's how it's been rendered by the chart. As clusterIP is an immutable value, Kubernetes rejects Helm's erroneous change and aborts the upgrade.

See https://github.com/helm/helm/issues/6378#issuecomment-557746499 for a more detailed explanation. Note that this does not affect Helm 2 as Helm 2's patching ignored the current state of the Kubernetes cluster (and therefore only saw a before value of empty string and a desired value of empty string, thus not generating any patch). Only Helm 3 is impacted due to Helm 3's three-way patching considering current cluster state as well.

This commit adds an if-block around the clusterIP definition in the Service template to ensure it is only included in the rendered output if it has been explicitly provided (and not just if it is set to empty string). Otherwise, its omission enables Kubernetes to continue to auto-assign an IP address upon installation and avoids an erroneous patch from being generated upon upgrades.


peimanja:
Any update on this one?

PR #1008, opened 2021-03-31
Bump velocity to 2.3

candrews:
Fixes CVE-2020-13936 / SNYK-JAVA-ORGAPACHEVELOCITY-1083992

The fix in Velocity is https://github.com/apache/velocity-engine/pull/16

Closes https://github.com/mock-server/mockserver/issues/1006

PR #1030, opened 2021-05-31
PRIO (Testfix included): CustomKeyAndCertificateFactorySupplier added

Arkinator:
## What problem is adressed?

I want to gain greater control over the SSL-Certificates used.

## What is the solution

Just provide a custom supplier for a custom implementation of a KeyAndCertificateFactory.

## What other solutions were discarded?

In keeping with Configuration-Approach more commonly used in Mockserver I first thought of introducing a new property for this case. Best I could come up with would be the class name, then just instantiate the given class in the KeyAndCertificateFactoryFactory, similiar to how this is done with BouncyCastle.
However this makes it impossible to have good control over the configuration of the Factory: The constructor will always be called only with MockserverLogger as the only argument. If there ever would be more then one MockServer in the same VM the use of constants to configure the instances of the Factories would not be possible anymore (and it would be a misleading design-choice to begin with)

## What else?

This PR solves the long persisting build issues by updating the certificates used in the tests. To make it easier for future-me I updated the MD describing the fix to include the critical udpate on the keychain.


candrews:
@jamesdbloom this PR fixes the longstanding test failures by the buildkite CI system which is causing all submitted PRs to fail, such as https://github.com/mock-server/mockserver/pull/1008

Arkinator:
Since I needed this PR I forked the project: https://search.maven.org/artifact/io.github.arkinator/mockserver/5.11.3/pom
I am not planning to upkeeping that repository and would love for this PR to be merged and the fork to die.

jamesdbloom:

Arkinator:
Wohoo! James is back! Will get on it :)

PR #1038, opened 2021-06-22
Just testing the CI

candrews:
If the buildkite tests fail for this PR, then clearly there's something wrong with those tests :)


jamesdbloom:
There was a bug in buildkite as for PR build some of the environment variables specifying folders were not correct.  I have updated the pipeline to work around this issue so I'm closing this PR.

PR #1048, opened 2021-07-14
Do not treat ndjson as json

tobiasstadler:
`org.mockserver.client.MockServerClient#retrieveRecordedRequests(org.mockserver.model.RequestDefinition)` returns only the first json object (instead of all) when the content type is `application/x-ndjson`, so do not treat `application/x-ndjson` as json.


tobiasstadler:
Thank You

PR #1063, opened 2021-08-20
Mockserver client to use JVM truststore by configuration.

brunopalos:
Hi,

This PR adds the JVM trust store to the MockServer client SSL context when configured to do so (e.g. `-Dmockserver.useJvmTrustCertificates="true"`).
This way one can avoid creating custom certificate chain file by using `mockserver.tlsMutualAuthenticationCertificateChain` when the certificates are already in the JVM trust store.

**Important note**
I'm unable to properly verify and run tests due to the following issue https://github.com/mock-server/mockserver/issues/1064 (master is broken).

### DIFF
--- a/jekyll-www.mock-server.com/mock_server/_includes/tls_configuration.html
+++ b/jekyll-www.mock-server.com/mock_server/_includes/tls_configuration.html
@@ -335,9 +335,9 @@
     <pre class="code" style="padding: 2px;"><code class="code">-Dmockserver.tlsMutualAuthenticationCertificateChain="/some/existing/path"</code></pre>
 </div>
 
-<button id="button_configuration_mockserver_client_use_jvm_truststore" class="accordion title"><strong>Inbound TLS X.509 Trusted Certificates (for Trusting Client X.509 Certificates)</strong></button>
+<button id="button_configuration_mockserver_client_use_jvm_truststore" class="accordion title"><strong>Inbound TLS X.509 JVM Trusted Certificates</strong></button>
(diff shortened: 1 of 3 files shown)


jamesdbloom:

brunopalos:
@jamesdbloom The rebase is done 👍

jamesdbloom:
@brunopalos   having looked at this closer I think it is already possible by setting `forwardProxyTLSX509CertificatesTrustManagerType` to `JVM` as described here: https://mock-server.com/mock_server/configuration_properties.html#button_configuration_trusted_tls_certs_group_for_proxied_reqs, can you please tell me what is missing from the existing logic?

brunopalos:
Hi @jamesdbloom,

Probably I wasn't explicit enough but I want to trust JVM certificates for the MockServer _Client_, not for the outbound requests from the server side. I have the MockServer _Server_ running behind a SSL termination with self signed certificates.

jamesdbloom:
Ok that makes sense, I take another look later now I understand the requirement.

jamesdbloom:
I have made this the default behaviour and commuted the change. I decided not to add a config value because there are already too many for TLS and it's hard for people to understand them already.

I think the change is in the latest SNAPSHOT release but can't remember if this ran since I commuted or not.

I'm going to close this PR as this behaviour is now supported and I could merge this due to too many other code changes making merge very difficult. Thank you for your contribution though as it made the implementation faster.

brunopalos:
Cool, thank you @jamesdbloom 👍

PR #1078, opened 2021-09-28
namespace values in helm NOTES.txt should use {{  .Release.Namespace }} rather than mockserver

JMCFTW:
There are part of namespace values in `helm/mockserver/templates/NOTES.txt`  is using `mockserver` rather than `{{ .Release.Namespace }}`


JMCFTW:
ping @jamesdbloom

PR #1090, opened 2021-10-31
Fix flaky in HttpResponseSerializerTest and HttpResponseDTOSerializerTest

CharlesZKQ:
org.mockserver.serialization.serializers.response.HttpResponseSerializerTest.shouldReturnFormattedResponseWithAllFieldsSet is flaky, which is the same reason with [1088](https://github.com/mock-server/mockserver/pull/1088). There is another test org.mockserver.serialization.serializers.response.HttpResponseDTOSerializerTest.shouldReturnFormattedResponseWithAllFieldsSet is also flaky, and could be solve by the same change.


jamesdbloom:
I've not seen these test fail due to the order of the fields changes across multiple machines (which are all Mac or Linux) but I do accept that it is very possible for that to have happened.  So I can understand better do you know under what circumstances the field order changed?

I have now added the following line `objectMapper.configure(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY, true)` to configure the creation of all ObjectMapper objects so I don't believe any tests should be flaky due to json field order.  Can you please recheck if you find the test flaky so I can close this PR as believe it is no longer needed.

jamesdbloom:
I'm going to close this PR because I assume the use of `objectMapper.configure(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY, true)` has resolved any field ordering issues

PR #1115, opened 2021-12-25
Upgrade to latest jackson library

tyge68:
It is often used, and so any code that would depend on a more recent version could be impacted.

I suggested to be up to date


jamesdbloom:
This has already been upgraded in a separate PR so closing here

Issue #1244, opened 2022-03-28
Don't use shaded dependencies by default


candrews:
**Describe the feature request**
Recently, there was a change to use shaded dependencies by default: mockserver-spring-test-listener

Shaded dependencies have a lot of disadvantages:
* They're not overridable, so if you need a different version of a shaded dependencies, you're out of luck
* They can conflict with the non-shaded variants if all classes/files/etc aren't renamed as shown by https://github.com/mock-server/mockserver/issues/1201
* They're redundant, causing more data to be transferred and jars to be bigger

**What you are trying to do**
Not use shaded jars by default.

**The solution you'd like**
Not use shaded jars by default. If a user wants a shaded jar, they can opt in by specifying the classifier.

**Describe alternatives you've considered**
Override the dependencies and remove the `shaded` classifier. That's workable, but I believe that's the wrong thing to do.

jamesdbloom:
I'm already in flight with a change to fix this.  The shaded jar was never intended to be used by `mockserver-spring-test-listener` that was just an error in the maven poms.  I'm in the process of fixing the shaded jar and working around limitation in maven such that only top level dependencies will all have a shaded jar that could be optionally used.  Sadly classifiers don't work for this despite the shaded plugin working that way due to limitations in maven and the understandable limitation of a single pom for all classifiers.

PR #1245, opened 2022-03-28
Don't use shaded dependencies by default

candrews:
Shaded dependencies can still be used on an opt-in basis by specified
the shaded classifier.

Closes: https://github.com/mock-server/mockserver/issues/1244

### DIFF
--- a/mockserver-examples/pom.xml
+++ b/mockserver-examples/pom.xml
@@ -86,12 +86,10 @@
         <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>mockserver-netty</artifactId>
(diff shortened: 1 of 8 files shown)


jamesdbloom:
I agree with the suggestion but I'm already in flight with a change to fix this.  The shaded jar was never intended to be used by `mockserver-spring-test-listener` that was just an error in the maven poms.  I'm in the process of fixing the shaded jar and working around limitation in maven such that only top level dependencies will all have a shaded jar that could be optionally used.  Sadly classifiers don't work for this despite the shaded plugin working that way due to limitations in maven and the understandable limitation of a single pom for all classifiers.

PR #1258, opened 2022-04-01
Add Bodys From File for JSON

albert0-hernandez:
Enabling to declare a file for the bodies of the request and/or response passing from this expectation:

```json
{
  "httpRequest": {
    "path": "/some/path/{cartId}/{maxItemCount}",
    "body": {
         "hello": "bye"
    }
  },
  "httpResponse": {
    "body": {
        "id": 0,
        "category": {
          "id": 0,
          "name": "string"
        },
        "name": "doggie",
        "photoUrls": [
          "string"
        ],
        "tags": [
          {
            "id": 0,
            "name": "string"
          }
        ],
        "status": "available"
     }    
  }
}
```
To this type of expectation:

```json
{
  "httpRequest": {
    "path": "/some/path/{cartId}/{maxItemCount}",
    "body": {
        "type" : "JSON",
        "jsonFile" : "/file/path/of/request_body.json.json",
        "contentType" : "application/json; charset=utf-8"
    }
  },
  "httpResponse": {
    "body": {
        "type" : "JSON",
        "jsonFile" : "/file/path/of/response_body.json.json",
        "contentType" : "application/json; charset=utf-8"
     }    
  }
}
```

### DIFF
--- a/mockserver-core/src/main/java/org/mockserver/model/JsonBody.java
+++ b/mockserver-core/src/main/java/org/mockserver/model/JsonBody.java
@@ -4,9 +4,12 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.io.Files;
 import org.mockserver.matchers.MatchType;
 import org.mockserver.serialization.ObjectMapperFactory;
+import org.apache.commons.lang3.StringUtils;
 
+import java.io.File;
 import java.nio.charset.Charset;
 import java.util.Arrays;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


jamesdbloom:
I'm not sure we should be adding this is a way that is specific to JSON.  What about other formats like XML or binary responses.   I think this needs to be modelled as a separate type of BodyWithContentType so it could cover more scenarios.  In addition this PR needs integration tests as unfortunately the PR currently is missing support for DTOs, dashboard, JSON schema and OpenAPI of the top of my head (and possibly some more).

albert0-hernandez:
Totally agree that this covered our current business case where the bodys are too large to maintain in the initializer file config.
Importing the body from a file will be more easily to maintain and less error-prone.

If you see that another approach is better to accept and add this feature I'm opened to upgrade the PR.

jamesdbloom:
This issue is on the backlog here: https://trello.com/c/67FmFwyv/189-request-and-response-body-from-file-path-filesystem-or-classpath and will likely be implemented in the next month or so.

One way to achieve this now would be to use a callback (class or method) and then inside the callback you can load any file content and use this to match the request body or return a response body.

jamesdbloom:
Make this change isn't massive but as the current release which will go out soon is focused on bugs I don't want to change the API so the release is quicker.  After this I could add functionality for file and release it in the SNAPSHOT.

jamesdbloom:
Closing this PR due to inactivity, item accepted only backlog and linked on backlog

PR #1291, opened 2022-04-16
Fix #1281: Fix the handling of "servers" section of OAS3 spec when building expectations.

### LINKED ISSUE #1281: "OpenAPISerialiser" not using the paths from the "servers" sections of OAS3 spec, when building expectations.
**Describe the issue**
The current implementation "OpenAPISerialiser" does not expect that OAS3 specification allows API spec designers to add paths to the server URLs in the "servers" sections of their specifications. As a result - the expectations created from OAS3 specs, which have such "servers" sections have partial endpoint paths and are not matching the incoming calls.

**What you are trying to do**
Modify the "OpenAPISerialiser" class to
* Check the "servers" sections at the 3 priority levels allowed by the OAS3 specification
  - HTTP operation level
  - HTTP path level
  - overall spec level
* Pick the correct "servers" section based on their scope and priority, when gathering the expectation data
* Pick the first URL from the "servers" section
* Substitute any variables used in the URL text with their "default" values provided in the OAS3 spec
* Extract the path fragment from the URL
* Append the path fragment as a prefix for the path of the prepared expectation

**MockServer version**
The version you are using (i.e. 5.13.2)

**To Reproduce**

* Use any OAS3 spec that has a "servers" section, where the URL has a path fragment.
* The created expectations will not include in their path the fragment from the "servers" section URL. 

**Expected behavior**
* The path fragments from the "servers" section URLs should be included as a prefix to the expectation path.

**MockServer Log**
Log output, as INFO level (or lower)

dkozhuharov:


dkozhuharov:
I'm stuck updating the tests in project `mockserver-integration-testing`. Help is highly appreciated 😄

jamesdbloom:
I found the following minor points
- I think not all the required tests were fixed by adding the "/v1" prefix.  
- I also found a few tests missing.  
- Also there is an existing class called URLParser that can return the path from a relative or fully qualified URL by using a simple regex.  This is well tested for all known URL formats.  The following code could be written which would remove the need for the URLSplitter class completely:

```java
URLParser.returnPath(getDefaultUrl(serverList.get(0)))
```

- I also saw some of the if-else conditions for string parsing like check if there is a "/" prefix can be replaced by single line StringUtils functions from commons-lang3.

I've made these changes and fixed the tests here: https://github.com/mock-server/mockserver/pull/1301. The build is running currently but I believe it should pass as all tests passed in IntelliJ and I'm half way through a full maven build locally.

I'm happy to either merged the PR (once the build passes) or for you fork the branch I created and create a new PR in your name as the majority of the hard parts came from you so the full PR including my bits could be merged in your name.  Let me know which you would prefer.

dkozhuharov:
RE: @jamesdbloom 
I like your rewrite 👌. Few notes on the testing:

I do have the test coverage for the "servers" section at the "path " level and at the "method" level.
If you check [openapi_petstore_example.json](https://github.com/mock-server/mockserver/pull/1291/files#diff-0a70b26e2ace77d0798e981926107dd1296228cd7ac42b13d39b61a7789d67ea) I've added "servers" sections under the path `/pets` and another under method `POST /pets`, which also has a variable in the URL.
So I changed the test correspondingly:
* tests that were hitting `GET /pets` were changed to expect `GET /v2/pets`
* tests that were hitting `POST /pets` were changed to expect `POST /v3/pets`
* tests that were hitting `GET /pets/{id}` were changed to expect `POST /v1/pets/{id}`
* tests that were hitting `GET /some/path` were changed to expect `POST /v1/some/path`

In your rewrite PR I saw a "servers" section under the path level, but I didn't see any added under the method level. It doesn't seem this case is covered in your PR.

It is o.k. if you merge your PR. Just link my PR in the description of your PR for traceability 👍

jamesdbloom:
Yes good point I hadn't adding that test.  I using your example json and add a test for that then merge the commit.

jamesdbloom:
I've covered that last scenario in my last commit so once the build passes I'll merge the PR

jamesdbloom:
Closes as changes now migrated to #1301

PR #1301, opened 2022-04-23
fixes #1281 support for server urls in OpenAPI specification, by adding server url path as path prefix to operations

### LINKED ISSUE #1281: "OpenAPISerialiser" not using the paths from the "servers" sections of OAS3 spec, when building expectations.

jamesdbloom:
fixes #1281 support for server urls in OpenAPI specification, by adding server url path as path prefix to operations

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
--
+- support for server urls in OpenAPI specification, by adding server url path as path prefix to operations
 
 ### Changed
 -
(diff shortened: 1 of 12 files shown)


jamesdbloom:
This PR is a modified version of PR #1291 so that PR for more discussions and background on this change.

dkozhuharov (commented):
Looks Good! 👍

PR #1327, opened 2022-05-20
Added Callback for the binary handler

Arkinator:
Added a callback to the binaryHandler. This enables traffic-capturing of non-http-traffic. To make this handling more meaningfull metadata was added to the exchange.

### DIFF
--- a/mockserver-core/src/main/java/org/mockserver/model/BinaryExchangeDescriptor.java
+++ b/mockserver-core/src/main/java/org/mockserver/model/BinaryExchangeDescriptor.java
@@ -1,24 +1,18 @@
 package org.mockserver.model;
 
-import java.net.InetSocketAddress;
-import java.time.LocalDateTime;
+import java.net.SocketAddress;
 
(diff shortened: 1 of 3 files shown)


jamesdbloom:
This looks good but:
- I think the dates should probably be pushed down into the BinaryMessage, say a receivedDateTime field.
- also I think it would be better to use SocketAddress for the type in the callback because InetSocketAddress could be null and this is not very clear
- with the changes about the callback could be call directly with three parameters response, response and socket, this would avoid the need for an extra class just to wrap those values

I think that would simplify the code and the way it would be used.  What do you think?

Arkinator:
fixed the first two findings.
The third is a matter of taste: I like avoiding long, ambiguous parameter lists. But for one: Netty just seems to love those. Plus this method is just called once, and this is internal...
Your call, I'll be happy to patch it out.

jamesdbloom:
It's not a biggy so I'm happy to merge this as is.  I agree its a matter of taste.

PR #1328, opened 2022-05-20
SSL-Context is now configureable independently for server and client

Arkinator:
To further enable customization of SSL-Contexts the factory now can distinguish if it is being asked to supply a context for a server or a client. This enables MockServer to present completely different Contexts for the server- and client-side. This is in addition to the customization in the configuration as it supplies more information to runtime customization (namely in the `KeyAndCertificateFactoryFactory.customKeyAndCertificateFactorySupplier`)


jamesdbloom:
This looks great but I think `isServerInstance` could be clearer if it was called `forServer`.  This is the terminology used in Netty and it seems slightly clearer.

Arkinator:
Thanks a bunch for the fast review. Rename added.

jamesdbloom:
Thanks I've merged this now

PR #1421, opened 2022-07-28
fix #1396 - relocate shaded com.google to avoid collisions at runtime

### LINKED ISSUE #1396: Using mockserver-client-java-no-dependencies causes package collision with com.google.common.cache
**Describe the issue**
When using Guava cache , and after replacing the `mockserver-client-java` dependency with `mockserver-client-java-no-dependencies` dependency we get the following error:
```
com.google.common.util.concurrent.ExecutionError: java.lang.NoSuchFieldError: ASCII
 	at com.google.common.cache.LocalCache$Segment.get(LocalCache.java:2053) ~[mockserver-client-java-no-dependencies-5.13.2.jar:na]
 	at com.google.common.cache.LocalCache.get(LocalCache.java:3966) ~[mockserver-client-java-no-dependencies-5.13.2.jar:na]
 	at com.google.common.cache.LocalCache.getOrLoad(LocalCache.java:3989) ~[mockserver-client-java-no-dependencies-5.13.2.jar:na]
 	at com.google.common.cache.LocalCache$LocalLoadingCache.get(LocalCache.java:4950) ~[mockserver-client-java-no-dependencies-5.13.2.jar:na]
 	at com.google.common.cache.LocalCache$LocalLoadingCache.getUnchecked(LocalCache.java:4956) ~[mockserver-client-java-no-dependencies-5.13.2.jar:na]
```

**Workaround**
I built `mockserver-client-java-no-dependencies:5.13.2` locally, with a minimal change to shading:
```
<relocation>
   <pattern>com.google</pattern>
   <shadedPattern>shaded_package.com.google</shadedPattern>
</relocation>
```
This seems to resolve the issue (The right jar is being used for the Guava package). I wonder why `com.google` is intentionally not relocated ?

**What you are trying to do**
I am implementing an integration test that involves a few test containers , including a mockserver test container and a mockserver client to configure it. We are using the `mockserver-client-java` dependency but it causes us a few problems with transitive dependencies so `mockserver-client-java-no-dependencies` seems like a natural solution to the problem and we would like to use it instead if possible. 

**MockServer version**
5.13.2

**To Reproduce**
1. Have a project with the following dependencies:
```
        <dependency>
            <groupId>com.google.guava</groupId>
            <artifactId>guava</artifactId>
            <version>19.0</version>
        </dependency>
        <dependency>
            <groupId>org.mock-server</groupId>
            <artifactId>mockserver-client-java-no-dependencies</artifactId>
            <version>5.13.2</version>
            <scope>test</scope>
        </dependency>
```

2. Create a Guava LoadingCache and use it.

kshlomi:
Proposal to fix #1396

### DIFF
--- a/mockserver-client-java/pom.xml
+++ b/mockserver-client-java/pom.xml
@@ -163,7 +163,10 @@
                                     <pattern>com.github</pattern>
                                     <shadedPattern>shaded_package.com.github</shadedPattern>
                                 </relocation>
-                                <!-- deliberately skipped com.google -->
+                                <relocation>
+                                    <pattern>com.google</pattern>
+                                    <shadedPattern>shaded_package.com.google</shadedPattern>
+                                </relocation>
                                 <relocation>
                                     <pattern>com.jayway</pattern>
                                     <shadedPattern>shaded_package.com.jayway</shadedPattern>


jamesdbloom:
I'm just testing against wider range of tests not yet integrated into the pipeline for docker and K8s.

I also noticed that you have only changes one instances of `<!-- deliberately skipped com.google -->` for the java client.  Can you please update the other 5 references unless this was deliberate.

jamesdbloom:
Actually maybe that is were the issue was, that com.google being skipped was required by the non clients, I think because the OpenAPI library was broken by the re-write.  In which case it would be safe to do this for the client but not the other dependencies.  I'll test this and if this is the case then I can merge your PR.

PR #1465, opened 2022-08-31
Fix references to globally-scoped values within Ingress template

KingJ:
343bded2c48857db9e042b4a99dcf95652024546 modified templating in a number of files, including the Ingress template, so that the release name could be overridden. This commit also changed templating of Ingress rules to be created for each host in the `ingress.hosts` values file using a `range` iterator. However, within a `range` iterator any `.` references refer to the iterator's scope (i.e `ingress.hosts`), not the global scope. As a result, references within this iterator (e.g to `ingress.path`) fail with an error. Within an iterator `$` can be used instead of `.` to refer to the global scope.

This PR adjusts the Ingress path templating to make use of the global scope again, as all values (except for `host`) are not part of the `ingress.hosts` scope.

This can be tested as follows using the following values file;

```
ingress:
  enabled: true
  hosts:
    - foo.bar
  path: /hello

```

When run against 5.14.0, the following error is returned;

```
helm template mockserver . -f values-test.yaml
Error: template: mockserver/templates/ingress.yaml:33:28: executing "mockserver/templates/ingress.yaml" at <.Values.ingress.path>: can't evaluate field Values in type interface {}
```

However, with this fix in place the Helm chart can be fully templated without error, with the Ingress section correctly templating the user-supplied values - the release name override introduced in 343bded2c48857db9e042b4a99dcf95652024546 is also present in the reference to the service name.

```
helm template mockserver . -f values-test.yaml
...
---
# Source: mockserver/templates/ingress.yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: mockserver
  labels:
    app: mockserver
    release: mockserver
    chart: mockserver-5.14.0
spec:
  rules:
    - host: foo.bar
      http:
        paths:
          - path: /hello
            backend:
              service:
                name: mockserver
                port:
                  number: 1080
            pathType: Prefix
---
...
```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -5,12 +5,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- fix references to globally-scoped values within Ingress template
 
 ### Added
 
 ### Changed
 - improved error message for not valid HTTP requests that are not being proxied
-- improved error message when client doesn't trust MockServer's CA 
+- improved error message when client doesn't trust MockServer's CA
 
 ### Fixed
 - fixed error passing configuration in MockServerClient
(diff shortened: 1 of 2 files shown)


KingJ:
It looks like my autoformatter trimmed trailing whitespace for a few older changelog entries - i'm happy to revert that commit and re-do it so it only has my entry, or I can leave it as-is.

PR #1475, opened 2022-09-07
Changed remoteAddress formatting

Arkinator:
In our project we need to access the TCP-Port of the client, from which the request originated. Rather then introducing yet another field I propose this more simplified change, still keeping in line with conventions of mockserver-addresses.


Arkinator:
This IS a breaking change, as reflected by the many unit-tests I had to touch. I feel this is a good and justified change, but that is a very subjective opinion. Please review carefully.
Also: I have no clue what that "lift" failure is about. It is complaining about general violations in the project which have nothing to do with this PR?!?!?

PR #1535, opened 2022-10-20
SSL fixes (tests green locally)

Arkinator:


Arkinator:
Fixes the "forServer" attribute (it wasn't forwarded in a crucial function) while also expanding the Supplier-Interface to include the configuration. This is necessary to e.g. create a BC-Fallback Factory

jamesdbloom:
Thanks for this PR

Arkinator:
Always a pleasure :)

PR #1565, opened 2022-11-25
Fix #1564. Ingress incorrect templating

### LINKED ISSUE #1564: Helm Install of 5.14.0 should take an array as an input for ingress path
Hello, I am installing the most recent version of the helm charts (5.14.0). Currently, we are passing in the same variables as 5.13.2, but encountering the error: 

`Error: UPGRADE FAILED: template: mockserver/templates/ingress.yaml:33:28: executing "mockserver/templates/ingress.yaml" at <.Values.ingress.path>: can't evaluate field Values in type interface {}`

The paths should read or process an array rather than a string value for the value of path, which is throwing the above error.

zifter:
Use default ingress template from helm chart example.

### DIFF
--- a/helm/mockserver/values.yaml
+++ b/helm/mockserver/values.yaml
@@ -33,17 +33,12 @@ ingress:
   enabled: false
   className: ""
   annotations: {}
-    # kubernetes.io/ingress.class: nginx
-    # kubernetes.io/tls-acme: "true"
   hosts:
-    - host: chart-example.local
+    - host: mockserver.local
       paths:
         - path: /
           pathType: ImplementationSpecific
   tls: []
-  #  - secretName: chart-example-tls
-  #    hosts:
-  #      - chart-example.local
 
 podAnnotations: {}
 
--- a/helm/mockserver/values.yaml
+++ b/helm/mockserver/values.yaml
@@ -30,16 +30,20 @@ service:
     image: radial/busyboxplus:curl
 
 ingress:
-  ingressClass:
-    enabled: false
-    name: ""
   enabled: false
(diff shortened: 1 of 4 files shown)


helm/mockserver/values.yaml
```diff
@@ -30,16 +30,20 @@ service:
...
 ingress:
-  ingressClass:
-    enabled: false
-    name: ""
   enabled: false
+  className: ""
   annotations: {}
-  path: /
-  pathType: Prefix
+    # kubernetes.io/ingress.class: nginx
```
jamesdbloom, line 36: Please remove these commented out lines

helm/mockserver/values.yaml
```diff
@@ -30,16 +30,20 @@ service:
...
+    # kubernetes.io/tls-acme: "true"
   hosts:
-    - mockserver.local
+    - host: chart-example.local
```
jamesdbloom, line 39: Please put back the default value as mockserver.local

helm/mockserver/values.yaml
```diff
@@ -30,16 +30,20 @@ service:
...
+      paths:
+        - path: /
+          pathType: ImplementationSpecific
   tls: []
+  #  - secretName: chart-example-tls
```
jamesdbloom, line 44: Please remove these commented out lines

zifter:
@jamesdbloom thanks for review! If will fix you review notes, but jsfyi this is how helm chart template looks like.
It's better to use default values from helm chart template, instead of modifying existing (i mean also to delete helpfull example comments and changing placeholder names).

PR #1614, opened 2023-01-04
DirectProxy asynchronous communication

Arkinator:
This PR focuses on asynchronous communication (non-paired messages). This is not part of http and is achieved by use of the BinaryRequestProxyingHandler, i.e. direct-proxy mode. Three main changes were made:
* An extra option forwardBinaryRequestsAsynchronously was added, enabling (or disabling, by default) the next two changes
* An extra mode was added to the NettyHttpClient. This mode keeps the socket to the remote server open. This is necessary when multiple packets from the client to the mockserver comprise one message. The mockserver should forward these messages using one socket channel.
* The binaryExchangeCallback is triggered immediately, not only after a response is received in the MockServer.


mockserver-core/src/main/java/org/mockserver/httpclient/NettyHttpClient.java
```diff
@@ -101,6 +102,7 @@ && isHostNotOnNoProxyHostList(remoteAddress)) {
                 .attr(SECURE, httpRequest.isSecure() != null && httpRequest.isSecure())
                 .attr(REMOTE_SOCKET, remoteAddress)
                 .attr(RESPONSE_FUTURE, responseFuture)
+                .attr(ENFORCE_SYNCHRONOUS_COMPLETION, true)
```
jamesdbloom, line 105: Won't this make asynchronous always on?

jamesdbloom: I guess in this case this attribute doesn't even need to be set, as long as it is queried expecting a null.

mockserver-netty/src/main/java/org/mockserver/netty/proxy/BinaryRequestProxyingHandler.java
```diff
@@ -123,7 +188,8 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
             mockServerLogger.logEvent(
                 new LogEntry()
                     .setLogLevel(Level.ERROR)
-                    .setMessageFormat("exception caught by " + this.getClass() + " handler -> closing pipeline " + ctx.channel())
+                    .setMessageFormat(
+                        "exception caught by " + this.getClass() + " handler -> closing pipeline " + ctx.channel())
```
jamesdbloom, line 192: Can you expand your line length setting so that these sorts of files don't get formatted onto one line.  Its not a biggy but its more consistent with the rest of the code if this type of line is not broken.

jamesdbloom:
@Arkinator thanks this looks interesting.  So I can understand what use case are you looking to solve for this?

I'm just adding support for HTTP2 and may have to do something similar later.  For now the HTTP2 support is fairly simple so it doesn't yet leave upstream connections open or support multiplexing.

mockserver-core/src/main/java/org/mockserver/httpclient/NettyHttpClient.java
Arkinator: I just set it for completeness. I dont want to design the use of the null-check in. This makes it explicit, which I like.

mockserver-netty/src/main/java/org/mockserver/netty/proxy/BinaryRequestProxyingHandler.java
Arkinator: ooopsie, will do

Arkinator:
Use case is the support of various protocols on our end which are non-http and do not follow a strict request/response flow. I originally wanted to support even more complex scenarios here (for example a classical web-socket flow with large messages should fail). But this would require a whole lot of rewrite in the httpClient-part. Basically you would need to synchronize the state of the up- and downstream socket, most importantly the closing. Since this was beyond my Netty-knowledge and our use-case was supported through this solution I was fine with leaving it as is.

mockserver-netty/src/main/java/org/mockserver/netty/proxy/BinaryRequestProxyingHandler.java
Arkinator: hmm, my setting is already at 120. what should i use?

jamesdbloom:
I've resolve the conflicts between a small recent changes and your PR, if the tests pass I'm happy to merge this.

jamesdbloom:
I'm just doing some work on HTTP2 then I'll do a release very soon

jamesdbloom:
@Arkinator Looks like your new tests are still failing.  I struggle to find time usually but if you would like me to take a look a can?

Arkinator:
that would be perfect :)
fixes pushed. have a great weekend

jamesdbloom:
I've now merged this.  Thank you for your helpful contribution this makes sense.

I am going to make some minor changes in the naming to improve the clarity and so that I can document this functionality for others to use and benefit from.  The essential the major or your code will be unchanged but words like asynchronous are not very clear but I believe for example `forwardBinaryRequestsWithoutWaitingForResponse` is much clearer for other people.

I'm also going to change the callback to be set via the configuration and to be a listener as this is both more consistent with MockServer style and not static which causes issues for people running MockServer multiple times in the same JVM.

These are all very small changes which I'll commit today or tomorrow.

Issue #1644, opened 2023-01-20
MemoryLeak in LRUCache


eric6iese:
**Describe the issue**
The field org.mockserver.cache.LRUCache.allCaches poses at least a memory leak, perhaps even a threadsafety issue.

**What you are trying to do**
We have a long running testsuite where we create and stop multiple ClientAndServer Instances in Java. There is never more than two or three active instances around, but the memory was increasing. In VisualVM i could dected that this class holds even the stopped HttpState-Objects in the LRUCache, which held a lot of LogEntry instances, even though the EventLog in MockServer was cleared. The reason for this is the internal disruptor which keeps hold of these instances - this is by design. 

The leak in LRUCache is the private static field allStates. It is used for nothing else than a static method LRUCache.clearAllCaches(), but it does not dected if those instances are even in use anywhere. Also I am not sure if this is even threadsafe, it isn't if multiple constructors of LRUCache could be called in parallel.

**MockServer version**
5.14, but I see no changes in the current implementation.

**To Reproduce**
Simple - create multiple ClientAndServer instances, do some interaction, stop them and watch what happens in the LRUCache with the debugger

**Fix **
The most simple fix would be to change the declaration to 

    private static final Set<LRUCache<?, ?>> allCaches = Collections.newSetFromMap(new WeakHashMap<>());

In addtion, synchronized blocks would be needed around the interactions.

asolntsev:
I see it's fixed by https://github.com/mock-server/mockserver/pull/1648
Is the release date known?

jamesdbloom:
This was already fixed in commit 6f143cf8e. The `allCaches` field was changed from a plain `List` to `Collections.synchronizedSet(Collections.newSetFromMap(new WeakHashMap<>()))`, exactly as you suggested. Once an `LRUCache` instance is no longer strongly referenced (e.g., after stopping a `ClientAndServer`), it becomes eligible for garbage collection and is automatically removed from the set.

This fix is in the current codebase and will be included in the next release.

PR #1648, opened 2023-01-23
#1644 Replaced the non-threadsafe List allCaches in the LRUCache

eric6iese:
with a synchronized weak HashSet. This prevents memory leaks with LRUCache-Instances which are no longer in use (and could hold onto suprisingly large disruptor-ringbuffers). It also prevents a possible (if unlikely) thread safety issues if multiple LRUCaches are created at the same time.

### DIFF
--- a/mockserver-core/src/main/java/org/mockserver/cache/LRUCache.java
+++ b/mockserver-core/src/main/java/org/mockserver/cache/LRUCache.java
@@ -1,22 +1,20 @@
 package org.mockserver.cache;
 
 import com.google.common.annotations.VisibleForTesting;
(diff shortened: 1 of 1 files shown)


jamesdbloom:
thanks, although this is only used during tests it is better to make if thread safe

asolntsev:
@eric6iese @jamesdbloom Is there a plan to release this fix?
We experience OutOfMemory problem in LittleProxy tests due to this memory leak:

https://github.com/LittleProxy/LittleProxy/actions/runs/7888633858

asolntsev:
@eric6iese @jamesdbloom Anyone? What about releasing this critical fix?

PR #1649, opened 2023-01-27
Bill Of Materials for Mock Server

tisoft:
Can be used to manage all artifact versions of MockServer in a `dependencyManagement`section of a pom.

```xml
    <dependencyManagement>
        <dependencies>
            <!-- mockserver -->
            <dependency>
                <groupId>org.mock-server</groupId>
                <artifactId>mockserver-bom</artifactId>
                <version>X.X.X</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>
        <dependencies>
    <dependencyManagement>
```

This makes it easier to have matching MockServer client/netty/junit dependencies in a project.


jamesdbloom:
This is already achieved using ‘ dependencyManagement’ in the parent pom.

Can you please explain how the BOM is different?

tisoft:
The dependency management of the parent pom is not complete, there are artifacts missing (e.g. mockserver-junit-rule). This could be fixed in the parent of course.

But the parent also manages all the dependencies of mock server itself and I don't want those managed dependencies to pollute my projects, specially if I use the `-no-dependencies`variants.

So the mockserver-bom only manages the mock server artifacts, not its dependencies.

(I tried to import the box in the parent, to remove the repeated management, but that created a cycle (of course) and doesn't work.)

mockserver-bom/pom.xml
```diff
@@ -0,0 +1,163 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
```
honnix, line 3: A bom should not have a parent that also manages dependencies, otherwise those dependencies managed by parent will be leaked through bom.

jamesdbloom:

**Rebased branch:** [`pr/1649-rebased`](https://github.com/mock-server/mockserver/tree/pr/1649-rebased)

jamesdbloom:
Thank you for this PR! A Maven BOM is a valuable contribution. However, I've implemented an **alternative approach using the `flatten-maven-plugin`** instead of a manually-maintained BOM module.

## What was implemented

**Commit:** https://github.com/mock-server/mockserver/commit/57ad0786ca27ebb30d5e85ecb25f31084e5b75f5

### Changes:

1. **Expanded parent POM `<dependencyManagement>`** to include all 22 published artifacts (up from 7):
   - All core modules
   - Shaded classifiers (`mockserver-client-java:shaded`, `mockserver-netty:shaded`, etc.)
   - `jar-with-dependencies` classifier (netty only)
   - `-no-dependencies` artifacts
   - WAR types and `classes` classifiers

2. **Added `flatten-maven-plugin` (1.6.0)** to auto-generate consumer-friendly BOM at build time

3. **Configured `.gitignore`** for generated `.flattened-pom.xml` file

### How consumers use it:

```xml
<dependencyManagement>
  <dependencies>
    <dependency>
      <groupId>org.mock-server</groupId>
      <artifactId>mockserver</artifactId>  <!-- import parent POM as BOM -->
      <version>5.15.1</version>
      <type>pom</type>
      <scope>import</scope>
    </dependency>
  </dependencies>
</dependencyManagement>

<dependencies>
  <!-- Now reference artifacts without versions -->
  <dependency>
    <groupId>org.mock-server</groupId>
    <artifactId>mockserver-netty</artifactId>
    <!-- Version inherited from BOM! -->
  </dependency>
</dependencies>
```

## Why flatten-maven-plugin instead of manual BOM?

### Your approach (manual BOM):
- ✅ Provides BOM functionality
- ❌ 163 lines of XML to maintain manually
- ❌ Must update when adding modules/classifiers
- ❌ Included non-existent artifacts (`mockserver-integration-testing:shaded`, `mockserver-integration-testing-no-dependencies`)
- ❌ Included internal artifacts (`mockserver-testing`, `mockserver-examples`)
- ❌ Element ordering issues (`<classifier>` before `<version>` in WAR entries)
- ❌ Will need restructuring during monorepo migration

### Flatten plugin approach:
- ✅ Zero maintenance - auto-syncs with modules
- ✅ Only publishes what actually exists
- ✅ Strips build configuration (plugins, properties, modules) for clean consumer POM
- ✅ Industry standard (Spring Boot, Micronaut, Jackson, Testcontainers use it)
- ✅ Survives monorepo migration (just move plugin config, BOM regenerates)
- ✅ No manual XML updates ever needed

## How flatten-maven-plugin works

At build time, it:
1. Reads parent `pom.xml` (with build config, properties, modules, dependencyManagement)
2. Strips out build-only elements (plugins, properties, modules, profiles)
3. Resolves `${project.version}` placeholders
4. Generates `.flattened-pom.xml` with only `<dependencyManagement>`
5. Publishes the clean version to Maven Central instead of full parent POM

**Result:** Consumers get a pure BOM without MockServer's build configuration.

## Strategic timing

MockServer is actively planning a **monorepo migration** (see `docs/plans/monorepo-migration.md`) that will consolidate 8 repositories:
- Maven modules will move to `mockserver/` subfolder
- All POMs will need path updates
- Manual BOM would require complete restructuring

The flatten plugin approach requires zero changes during monorepo - just move the plugin config and it continues working.

## Closing this PR

I'm closing this PR in favor of the flatten plugin implementation. The functionality you requested is now available in master, but with automated maintenance instead of manual upkeep.

Thank you again for identifying the need for a BOM and for the detailed PR! The flatten plugin solution builds on your idea while eliminating the maintenance burden.

If you have any questions about using the new BOM, please let me know!

jamesdbloom:
Closing in favor of flatten-maven-plugin implementation (commit 57ad0786c). The BOM functionality is now available with zero maintenance. Thank you for the contribution!

PR #1657, opened 2023-02-02
Remove implicit reliance on internal java-certificate-classes

Arkinator:
Deprecated classes (X509V1CertImpl) are implicitly relied upon by this code. This lead to a funny behaviour when inadvertently running the application on Java 17: TLS was sucessfully established, but certificate details were not retrievable. This fix relies on BouncyCastle instead, giving a more robust base for people using more modern JVMs.
The catch-clause was also cleaned up (InterruptedException will not be swallowed anymore)


jamesdbloom:
Thanks this has now be implemented but due to issues with the build I couldn't merge this PR.  I referenced you name in the commit though.

PR #1729, opened 2023-04-22
fix proactivelyInitialiseTLS example in tls-configuration doc

DenilssonMontoya:
Hi, 
Browsing documentation I noticed that there is an example of the "Proactively Initialise TLS During Start Up" setting in the HTTPS & TLS documentation that is not accurate.

Current example is as follow:
-Dmockserver.proactivelyInitialiseTLS="**/some/existing/path**"

Changed to:
-Dmockserver.proactivelyInitialiseTLS="**true**"

This setting is boolean type so it expect a true/false value.

### DIFF
--- a/jekyll-www.mock-server.com/mock_server/_includes/tls_configuration.html
+++ b/jekyll-www.mock-server.com/mock_server/_includes/tls_configuration.html
@@ -62,7 +62,7 @@
     <p>Property File:</p>
     <pre class="code" style="padding: 2px;"><code class="code">mockserver.proactivelyInitialiseTLS=...</code></pre>
     <p>Example:</p>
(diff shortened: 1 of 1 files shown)


jamesdbloom:

**Rebased branch:** [`pr/1729-rebased`](https://github.com/mock-server/mockserver/tree/pr/1729-rebased)

jamesdbloom:
✅ This fix has been merged in commit 7dc197541

You're absolutely right - `proactivelyInitialiseTLS` is a boolean property, not a file path. The documentation example has been corrected to show `"true"` instead of `"/some/existing/path"`.

Thank you for catching this documentation error! 🙏

PR #1755, opened 2023-06-12
Preserve exact bytes even when encoding is off

Arkinator:
We have seen this behavior when decoding MTOM-Request. MTOM is primarily used for non-conformant binary bodies, so the body can NOT be accurately recomposed from the string representation. This is fixed.
This is however a fix with a broader reach. In my opinion mockserver should always strive to alter the messages as little as possible (as it does for every other content-type). This improves on this metric (and judging from the other cases, it was just an oversight).

### DIFF
--- a/mockserver-core/src/main/java/org/mockserver/codec/BodyDecoderEncoder.java
+++ b/mockserver-core/src/main/java/org/mockserver/codec/BodyDecoderEncoder.java
@@ -91,6 +91,7 @@ public class BodyDecoderEncoder {
             } else if (mediaType.isXml()) {
                 return new XmlBody(
                     new String(bodyBytes, mediaType.getCharsetOrDefault()),
+                    bodyBytes,
                     mediaType
                 );
             } else if (mediaType.isString()) {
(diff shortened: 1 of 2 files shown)


jamesdbloom:

**Rebased branch:** [`pr/1755-rebased`](https://github.com/mock-server/mockserver/tree/pr/1755-rebased)

jamesdbloom:
## Merged with Thanks

Thank you for this excellent bug fix! I've merged this PR to preserve raw bytes for XML bodies.

### Problem

When decoding XML bodies, MockServer was not preserving the original byte array, unlike JSON and String bodies. This caused data corruption when:

1. **MTOM attachments** - Binary data embedded in XML/SOAP requests
2. **Charset mismatches** - When declared encoding doesn't match actual bytes
3. **Non-conformant binary content** - Binary data that cannot be accurately represented as strings

Without the raw bytes,  reconstructed bytes from the string representation, which altered binary data during lossy charset conversions.

### Solution

Now  receives and preserves the original  parameter, exactly like  and  already do. This ensures MockServer alters proxied messages as little as possible.

### Testing

✅ Added test `shouldNotAlterBodyForXmlsWithWrongCharset` that verifies raw bytes are preserved
✅ All existing tests pass
✅ Verified consistency with JSON/String body handling

Merged in commit: https://github.com/mock-server/mockserver/commit/4ff56dfb785b6934e080bcd8e9c444116ce5ee2c

PR #1774, opened 2023-07-18
Bugfix; velocity usage is not threadsafe

dom54:
ToolContext is not threadsafe, you must create a new context with every usage.  Not sure if this is a velocity bug as it implies putting tools in "request" scope means they will be recreated, but they are not.

### DIFF
--- a/mockserver-core/src/main/java/org/mockserver/templates/engine/velocity/VelocityTemplateEngine.java
+++ b/mockserver-core/src/main/java/org/mockserver/templates/engine/velocity/VelocityTemplateEngine.java
@@ -6,7 +6,6 @@ import com.google.common.base.Splitter;
 import org.apache.velocity.VelocityContext;
 import org.apache.velocity.app.VelocityEngine;
 import org.apache.velocity.runtime.RuntimeConstants;
-import org.apache.velocity.tools.ToolContext;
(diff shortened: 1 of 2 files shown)


dom54:
@jamesdbloom  any chance you can give me write access?

jamesdbloom:

**Rebased branch:** [`pr/1774-rebased`](https://github.com/mock-server/mockserver/tree/pr/1774-rebased)

jamesdbloom:
Merged in commits fb0c29845 and cccbf7b4c.

This PR fixes a thread-safety bug in the Velocity template engine.

### The Problem:

`ToolContext` is not thread-safe, but the code was storing a shared `ToolContext` instance as a field and reusing it across all template executions. This caused race conditions when multiple threads used Velocity templates concurrently, particularly with request-scoped tools like `$json`.

### The Fix:

- Store `ToolManager` (thread-safe) instead of `ToolContext` (not thread-safe) as instance variable
- Call `toolManager.createContext()` for each template execution to get a fresh `ToolContext`
- Added test `shouldUseRequestScopeToolsInThreadSafeWay` that verifies 100 concurrent requests with `$json.parse()` all get the correct isolated state

Thank you @dom54 for finding and fixing this concurrency bug! 🎉

Issue #1779, opened 2023-07-21
Use Persistent Volume Claim (PVC) in Helm Chart


MitchelNijdam-Rockstars:
**Describe the feature request**

MockServer has the ability to persistently save any expectation into a file. When running MockServer on Kubernetes, since pods are ephemeral and clear any files on restart - it would be nice to define a PVC to persistently store the configuration.

**What you are trying to do**

Keep MockServer's expectations persisted on Kubernetes pod restart.

**The solution you'd like**

* New value in `values.yaml`: `app.persistentVolumeClaimName`
* Update `templates/deployment.yaml` - if variable is defined, add volume & volumeMount
* Add extra template for PVC - surrounded with `{{ if .Values.app.persistentVolumeClaimName }}`

**Describe alternatives you've considered**

Current alternative is manually copy-pasting the whole helm chart to our gitops repository and modifying above manually. Would be nice if it became available for everyone.

jamesdbloom:
Implemented in commit 1d0bdbf85.

The Helm chart now supports opt-in PersistentVolumeClaim storage for persisting expectations across pod restarts. Enable it with:

```bash
helm upgrade --install --namespace mockserver \\
  --set app.persistence.enabled=true \\
  mockserver helm/mockserver
```

**What's included:**
- New `app.persistence` values block with 7 configurable options (`enabled`, `existingClaimName`, `storageClass`, `accessModes`, `size`, `mountPath`, `annotations`)
- Auto-configuration of `MOCKSERVER_PERSIST_EXPECTATIONS`, `MOCKSERVER_PERSISTED_EXPECTATIONS_PATH`, and `MOCKSERVER_INITIALIZATION_JSON_PATH` environment variables (safe defaults — any matching property in your `mockserver.properties` takes precedence)
- Support for both chart-managed PVCs and referencing existing PVCs
- Full documentation added to the Helm chart docs, Persisting Expectations page, internal docs, and in-chart README
- CI validation for persistence template rendering

**Backward compatible:** Disabled by default (`app.persistence.enabled: false`). Existing deployments are unaffected.

PR #1780, opened 2023-07-21
add custom PVC option to helm chart

MitchelNijdam-Rockstars:
Relates to https://github.com/mock-server/mockserver/issues/1779

This allows users to persist Expectation across pod restarts on Kubernetes. 

With this change, users can create their own PVC and attach it to the pod using the Helm chart values.


MitchelNijdam-Rockstars:
Closing as there is apparently no interest in this

Issue #1524, opened 2022-10-11
MockServerClient.verify(RequestDefinition requestDefinition, VerificationTimes times) not works correctly after update to 5.14.0


dorohovichandrey:
**Describe the issue**
After the update from 5.13.2 to 5.14.0 MockServerClient.verify(RequestDefinition requestDefinition, VerificationTimes times) start working not correctly in the cases, where expectations were defined with Path Parameters

**What you are trying to do**
https://github.com/dorohovichandrey/mockserver-example/blob/5.13.2/src/test/kotlin/com/example/MockserverVerifyTests.kt - tests passed on MockServer version 5.13.2

https://github.com/dorohovichandrey/mockserver-example/blob/5.14.0/src/test/kotlin/com/example/MockserverVerifyTests.kt - tests failed on MockServer version 5.14.0

**To Reproduce**
Steps to reproduce the issue:
1. define expectation with path parameter
2. call mocked endpoint
3. invoke MockServerClient.verify(RequestDefinition requestDefinition, VerificationTimes times)
https://github.com/dorohovichandrey/mockserver-example/blob/5.14.0/src/test/kotlin/com/example/MockserverVerifyTests.kt 

**How you are running MockServer (i.e maven plugin, docker, etc)**
In Docker by the Testcontainers (tested in both ARM and AMD envs)

**What error you saw**
```
Request not found exactly once, expected:<{
  "method" : "GET",
  "path" : "/something/{id}"
}> but was:<{
  "headers" : {
    "content-length" : [ "0" ],
    "Accept" : [ "*/*" ],
    "content-encoding" : [ ".*" ],
    "Connection" : [ "Keep-Alive" ],
    "User-Agent" : [ "Apache-HttpClient/4.5.13 (Java/17.0.3)" ],
    "Host" : [ "localhost:60503" ],
    "Accept-Encoding" : [ "gzip,deflate" ]
  },
  "keepAlive" : true,
  "method" : "GET",
  "path" : "/something/123",
  "secure" : false
}>
<Click to see difference>

java.lang.AssertionError: Request not found exactly once, expected:<{
  "method" : "GET",
  "path" : "/something/{id}"
}> but was:<{
  "headers" : {
    "content-length" : [ "0" ],
    "Accept" : [ "*/*" ],
    "content-encoding" : [ ".*" ],
    "Connection" : [ "Keep-Alive" ],
[...]

Issue #1505, opened 2022-09-26
Regexp matcher Illegal repetition


olivierACRI:
**Describe the issue**
Using regexp the mockserver is throwing an exception

**What you are trying to do**
Trying to catch all the request matching a certain regexp pattern

**MockServer version**
5.14.0

**To Reproduce**
Create an expectation as describe below 
Shoot the mockserver with the GET http://URL_OF_THE_SERVER/Batches(6f1298ed-68e5-4f5a-921b-17f64d9cc8d8)

1. How you are running MockServer (i.e maven plugin, docker, etc) :Docker
2. Code you used to create expectations
```
 {
        "id": "s2gm_get_single_order_mock",
        "httpRequest": {
          "method": "GET",
          "path": "/Batches({jobId})",
          "pathParameters": {
            "jobId": ["[a-z0-9\\-]+"]
          }
        },
        "httpResponse": {
          "body": {
            "BatchId": "{{ request.pathParameters.jobId.0 }}",
            "status": 'Done'
          }
        }
      },
```
5. What error you saw
```
[...]

PR #1791, opened 2023-08-29
Fix path parameter invalid pattern - issues: #1524, 1505

szada92:
Fixes issues:
- https://github.com/mock-server/mockserver/issues/1505
- https://github.com/mock-server/mockserver/issues/1524

During verification, `HttpRequestPropertiesMatcher` calls `matches` function to verify the RequestDefinition against itself - resulting the same behaviour as can be seen in newly added test, that the path parameter name will be added also as a value to the MultiValueMap. 
With this modification `PathParametersDecoder#extractPathParameters` method will not alter the stored RequestDefinition, and will not result the invalid Pattern matching for path parameter name Strings, which are not indended to be valid regex patterns.

### DIFF
--- a/mockserver-core/src/main/java/org/mockserver/codec/PathParametersDecoder.java
+++ b/mockserver-core/src/main/java/org/mockserver/codec/PathParametersDecoder.java
@@ -67,7 +67,7 @@ public class PathParametersDecoder {
     }
 
     public Parameters extractPathParameters(HttpRequest matcher, HttpRequest matched) {
-        Parameters parsedParameters = matched.getPathParameters() != null ? matched.getPathParameters() : new Parameters();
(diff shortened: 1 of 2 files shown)


jamesdbloom:

**Rebased branch:** [`pr/1791-rebased`](https://github.com/mock-server/mockserver/tree/pr/1791-rebased)

jamesdbloom:
Merged via rebased branch `pr/1791-rebased` in commit 34e62acc4.

This PR fixes a mutation bug where `extractPathParameters()` was modifying the original `HttpRequest` object's path parameters, causing:
- **Issue #1505**: Regex pattern compilation errors (`Illegal repetition`)
- **Issue #1524**: Verification failures after upgrading to 5.14.0

The fix: Clone the `Parameters` object before mutation instead of using a direct reference.

Thank you @szada92 for identifying and fixing this bug! 🎉

Issue #1834, opened 2024-01-24
Multiple responses for the same request are not thread-safe


dzianis-kryvapusk:
**Describe the issue**
I noticed that MockServerClient responds with only the first mocked response even if it has multiple of them when an app calls mocked API concurrently many times.

**What you are trying to do**
An app calls third-party API many times in a non-blocking manner. It means that MockServerClient gets many requests at the same time. This third-party API should respond with a different response every time. I mock them but receive only the first mocked response. When I added a delay between calls, everything began to work as expected.

**MockServer version**
org.mock-server:mockserver-spring-test-listener:5.15.0

**To Reproduce**
I prepared a [project](https://github.com/dzianis-kryvapusk/simple-spring-boot-async-microservice) with a [test (test_getJokes_successful)](https://github.com/dzianis-kryvapusk/simple-spring-boot-async-microservice/blob/a7490259a9108b1b65c318770e3fcd27d657784a/src/test/java/com/example/jokemicroservice/controller/JokeControllerTest.java#L61) which you can run and see the issue. You will find that the response from the server contains only the first mocked response repeated many times. That's why `assertEquals(expected, response)` will fail. To fix it, you need to uncomment lines in the method [com.example.jokemicroservice.jokeprovider.ThirdPartyJokeProvider#getJokesAsync](https://github.com/dzianis-kryvapusk/simple-spring-boot-async-microservice/blob/1e0fe0db5bd923b34a548196e5bdde219c99b74a/src/main/java/com/example/jokemicroservice/jokeprovider/ThirdPartyJokeProvider.java#L60).
You can run this test directly from your IDE.
Expectations are built by the following code:
```
for (int i = 1; i <= 5; i++) {
            JokeDto joke = buildJoke(i);
            expected.add(joke);
            mockServer.when(request()
                                    .withMethod(HttpMethod.GET.name())
                                    .withPath(jokeApiProperties.getPath()),
                            once())
                    .respond(response()
                                     .withStatusCode(HttpStatusCode.OK_200.code())
                                     .withContentType(org.mockserver.model.MediaType.APPLICATION_JSON)
                                     .withBody(objectMapper.writeValueAsString(joke)));
        }
```

**Expected behaviour**
The server should respond with a mocked response only once even on concurrent calls.

**MockServer Log**
In the attached file you can find many logs like
```
matched expectation:

  {
    "httpRequest" : {
      "method" : "GET",
      "path" : "/random_joke"
    },
[...]

PR #1835, opened 2024-01-24
fix some unit tests in mockserver-core

dzianis-kryvapusk:
I wanted to investigate [the issue](https://github.com/mock-server/mockserver/issues/1834) that I found yesterday but couldn't build the project because I had many failed tests. I run it on Windows and understand that it's one of the problems but I saw also many problems not related to the OS.
There are many non-fixed tests left, I just don't have time to fix all of them.
@jamesdbloom I would be glad if you find my contribution useful.


dzianis-kryvapusk:
looks like I need to revert some changes because it works differently on my local machine.

PR #1975, opened 2026-01-22
not only filter out optional strings but also notted strings

### LINKED ISSUE #1974: NottableString not does not work correctly when there are matches.
## Bug Description
Because it is currently not possible to strictly match only query parameters in the expectation (see: https://github.com/mock-server/mockserver/issues/1879) the next best thing is to use `NottableString.not` on the parameters that should not be present. But this seems to not work correctly.

**What I am trying to do**
I want to use `NottableString.not` to add make sure that the expectation only passes, when certain parameter names are not used in a request. But the rules for when they work are really strange. When no query parameter is added to the request I can add as many not requirements to the expectation as I want. But when adding a parameter to the request, suddenly only one not requirement is allowed (of course with another parameter name notted, than was used in the request).

What also does not work is if I want to explicitly check that one parameter is present but another is not.

It seems to me that the moment multiple `withQueryStringParameter` are used nottable parameters stop working.

**MockServer version**
5.15

## To Reproduce
Steps to reproduce the issue:

I use MockServer via the junit dependency (https://central.sonatype.com/artifact/org.mock-server/mockserver-junit-jupiter/overview)
The following Code can be used to reproduce the weird behaviour:

Code to test:

```java
import org.springframework.stereotype.Service;
import org.springframework.web.client.RestClient;

@Service
public class RestService {

	public String response(String url) {
		return RestClient.builder()
				.baseUrl(url)
				.build()
				.get()
				.uri(uriBuilder -> uriBuilder.path("/test")
						.queryParam("name", "John")
						.build())
				.retrieve()
				.body(String.class);
	}
}
```

In the code above only one query parameter is used (name) and no other parameter is used additionally. So the below test should pass. It expects that the name parameter is present and the age parameter is not present. This is of course the case. But the test still fails.

Test that should pass:

```java
import org.junit.jupiter.api.BeforeAll;
import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.extension.ExtendWith;
import org.mockito.InjectMocks;
import org.mockito.junit.jupiter.MockitoExtension;
import org.mockserver.client.MockServerClient;
import org.mockserver.junit.jupiter.MockServerExtension;
import org.mockserver.model.HttpRequest;
import org.mockserver.model.HttpResponse;

import static org.assertj.core.api.Assertions.assertThat;
import static org.mockserver.model.NottableString.not;

@ExtendWith({MockServerExtension.class, MockitoExtension.class})
class RestServiceTest {

	private static MockServerClient msc;

	@InjectMocks
	private RestService restService;

	@BeforeAll
	public static void initMockserver(MockServerClient client) {
		RestServiceTest.msc = client;
	}

	@Test
	void response() {
		msc
				.when(HttpRequest.request()
						.withQueryStringParameter("name", "John")
						.withQueryStringParameter(not("age"))
						.withMethod("GET"))
				.respond(HttpResponse.response()
						.withBody("Hello World!"));

		String response = restService.response("http://localhost:" + msc.getPort());

		assertThat(response).isEqualTo("Hello World!");
	}
}
```

The test can be “fixed” by either keeping only the positive requirement for name. or only using the negative requirement for age.

**Expected behaviour**
The test passes.

**MockServer Log**

```
18:58:27.066 [MockServer-EventLog0] INFO org.mockserver.log.MockServerEventLog -- 63852 request:

  {
    "method" : "GET",
    "path" : "/test",
    "headers" : {
      "content-length" : [ "0" ],
      "User-Agent" : [ "Java-http-client/25.0.1" ],
      "Upgrade" : [ "h2c" ],
      "Host" : [ "localhost:63852" ],
      "HTTP2-Settings" : [ "AAEAAEAAAAIAAAAAAAMAAAAAAAQBAAAAAAUAAEAAAAYABgAA" ],
      "Connection" : [ "Upgrade, HTTP2-Settings" ],
      "Accept-Encoding" : [ "gzip", "deflate" ]
    },
    "keepAlive" : true,
    "secure" : false,
    "protocol" : "HTTP_1_1",
    "localAddress" : "127.0.0.1:63852",
    "remoteAddress" : "127.0.0.1:63854"
  }

 didn't match expectation:

  {
    "httpRequest" : {
      "method" : "GET",
      "queryStringParameters" : {
        "name" : [ "John" ],
        "!age" : [ ".*" ]
      }
    },
    "httpResponse" : {
      "body" : "Hello World!"
    },
    "id" : "2e092b71-44a1-4dde-8eed-1b4f6a227c00",
    "priority" : 0,
    "timeToLive" : {
      "unlimited" : true
    },
    "times" : {
      "unlimited" : true
    }
  }

 because:

  method matched
  path matched
  body matched
  headers matched
  cookies matched
  pathParameters matched
  queryParameters didn't match
```

## Additional Context
Digging deeper we found out that there is a special case for when there are no parameters at all (`matched.isEmpty()`). Then the test passes if all the requirements are either notted or optional.

https://github.com/mock-server/mockserver/blob/master/mockserver-core/src/main/java/org/mockserver/matchers/MultiValueMapMatcher.java#L42-L48

But if some parameters are sent in the request another codepath is used. In this code path the notted requirements are not correctly taken into account. But it seems to be that the optionals are taken into account.

https://github.com/mock-server/mockserver/blob/master/mockserver-core/src/main/java/org/mockserver/collections/SubSetMatcher.java#L31-L33

We are pretty sure that there is a bug on line 31. Shouldn’t the following code be used instead?

```java
long subsetNonOptionalSize = subset.stream()
    .filter(ImmutableEntry::isNotOptional)
    .filter(entry -> !entry.isNotted())
    .count();
```

Notice the additional filter that takes care, that nottable parameters are not added to the total count.

A better name for the variable would probably be `subsetPositiveMatchesSize` or something similar.

DarioViva42:
fixes https://github.com/mock-server/mockserver/issues/1974

Some small additions to also consider.
Maybe add some tests that check if now mixing different type of strings (nottable & optional & ‘normal’) works correctly.

Also what I find kind of strange is that here https://github.com/mock-server/mockserver/blob/master/mockserver-core/src/main/java/org/mockserver/collections/ImmutableEntry.java#L36-L42 isOptional only checks that key is optional. But isNotted not only checks the key but also that the value is not notted. Why the difference?

### DIFF
--- a/mockserver-core/src/main/java/org/mockserver/collections/ImmutableEntry.java
+++ b/mockserver-core/src/main/java/org/mockserver/collections/ImmutableEntry.java
@@ -45,6 +45,10 @@ public class ImmutableEntry extends Pair<NottableString, NottableString> impleme
         return !isOptional();
     }
 
+    public boolean isNotNotted() {
+        return !isNotted();
+    }
+
     @Override
     public NottableString getLeft() {
         return key;
--- a/mockserver-core/src/main/java/org/mockserver/collections/SubSetMatcher.java
+++ b/mockserver-core/src/main/java/org/mockserver/collections/SubSetMatcher.java
@@ -28,9 +28,12 @@ public class SubSetMatcher {
         }
 
         if (result) {
-            long subsetNonOptionalSize = subset.stream().filter(ImmutableEntry::isNotOptional).count();
+            long subsetRequiredSize = subset.stream()
+                .filter(ImmutableEntry::isNotOptional)
+                .filter(ImmutableEntry::isNotNotted)
+                .count();
             // this prevents multiple items in the subset from being matched by a single item in the superset
-            result = matchingIndexes.size() >= subsetNonOptionalSize;
+            result = matchingIndexes.size() >= subsetRequiredSize;
         }
         return result;
     }
--- a/mockserver-core/src/test/java/org/mockserver/collections/SubSetMatcherTest.java
+++ b/mockserver-core/src/test/java/org/mockserver/collections/SubSetMatcherTest.java
@@ -313,4 +313,74 @@ public class SubSetMatcherTest {
         );
     }
 
+    @Test
+    public void shouldContainSubsetWithPositiveAndNottedKeys() {
+        // Issue #1974: notted parameters should not count toward required matches
+        // Expectation: name=John AND NOT age
+        // Request: name=John (no age parameter)
+        // Should match because: 1 positive match (name) >= 1 required (name), and age is not present
+        assertTrue(containsSubset(null, null, regexStringMatcher,
+            Arrays.asList(
+                new ImmutableEntry(regexStringMatcher, "name", "John"),
+                new ImmutableEntry(regexStringMatcher, "!age", ".*")
+            ),
+            new ArrayList<>(Arrays.asList(
+                new ImmutableEntry(regexStringMatcher, "name", "John")
+            )))
+        );
+    }
+
+    @Test
+    public void shouldContainSubsetWithMultiplePositiveAndNottedKeys() {
+        // Expectation: name=John AND city=London AND NOT age AND NOT country
+        // Request: name=John, city=London (no age or country)
+        // Should match because: 2 positive matches >= 2 required, and age/country not present
+        assertTrue(containsSubset(null, null, regexStringMatcher,
+            Arrays.asList(
+                new ImmutableEntry(regexStringMatcher, "name", "John"),
+                new ImmutableEntry(regexStringMatcher, "city", "London"),
+                new ImmutableEntry(regexStringMatcher, "!age", ".*"),
+                new ImmutableEntry(regexStringMatcher, "!country", ".*")
+            ),
+            new ArrayList<>(Arrays.asList(
+                new ImmutableEntry(regexStringMatcher, "name", "John"),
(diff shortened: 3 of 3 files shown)


jamesdbloom:

**Rebased branch:** [`pr/1975-rebased`](https://github.com/mock-server/mockserver/tree/pr/1975-rebased)

jamesdbloom:
Merged in commits 855719173 and b0942e216.

This PR fixes the notted query parameter matching bug reported in issue #1974.

### The Problem

When using `NottableString.not()` to assert that certain parameters should NOT be present, the matcher was incorrectly counting notted entries toward the required match count. This caused false negatives when mixing positive and notted parameters.

For example, this expectation should match a request with `?name=John` but was failing:
```java
.withQueryStringParameter("name", "John")
.withQueryStringParameter(not("age"))
```

### The Fix

Modified `SubSetMatcher` to exclude notted entries from the required match count using the new `isNotNotted()` helper method. Notted entries are negative assertions - they verify that certain keys are NOT present, handled separately by the `nottedAndPresent` check.

Added comprehensive test coverage for mixed positive/notted parameters, multiple notted parameters, optional + notted combinations, and cases where notted parameters ARE present.

Thank you @DarioViva42 for identifying and fixing this bug!

Closes #1974

PR #2459, opened 2026-07-26
fix(testcontainers-python): read exposed ports through an int normaliser

jamesdbloom:
## Why

`master` is red. The `MockServer Python` pipeline has failed on every build since #618, and because the umbrella `MockServer` pipeline triggers it, the umbrella build is red too (#6237–#6241 all failed for this single reason).

Root cause is an upstream contract change, not a MockServer regression: **testcontainers 4.15.0** keys `DockerContainer.ports` by `str(port)` rather than `int`.

```python
def with_exposed_ports(self, *ports: Union[str, int]) -> Self:
    for port in ports:
        self.ports[str(port)] = None   # <- stringified
    return self
```

and the attribute is now typed `dict[str, Optional[int]]`. The suite asserted `int` membership, so it started failing with:

```
AssertionError: assert 1080 in {'1080': None}
```

5 failed, 17 passed, 4 deselected.

## What changed

Tests only. Exposed ports are now read through a normaliser that parses each key to an `int` (tolerating a `"1080/tcp"` protocol suffix), so the assertions hold under either key style and the `testcontainers>=4.0.0` floor stays honest.

`MockServerContainer` itself was already correct and is **unchanged** — `get_exposed_port` takes an `int` and normalises internally, so the runtime path never broke.

## Latent false green also closed

After the keys became strings, this assertion in `test_replaces_default_port` passed *trivially*:

```python
assert MOCKSERVER_PORT not in container.ports   # 1080 not in {'1080': None} -> always True
```

It would no longer have caught `with_server_port` failing to drop the previously exposed port. Verified by degrading `with_server_port` to skip its `self.ports = {}` reset and confirming the test now fails:

```
assert 1080 not in {1080, 9090}
```

then restoring and re-running green.

## Verification

Run locally against testcontainers 4.15.0 with Docker Desktop 29.5.3 up:

- CI's exact command `pytest -m "not docker" -v` — **22 passed, 4 deselected** (was 5 failed, 17 passed)
- Docker-gated tests `pytest -m "docker"` — **4 passed** (these start a real MockServer container)
- Degrade-and-confirm-red as above, to prove the assertions still bite

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -244,7 +244,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `rate_limit_error` envelope with the exhausted rate-limit headers and `Retry-After`.
 
 ### Fixed
-- **BREAKING BEHAVIOUR: expectations persisted to a non-filesystem blob store are now saved under
+- **The `testcontainers-mockserver` (Python) port assertions no longer break against testcontainers
+  4.15.0, which keys `DockerContainer.ports` by `str(port)` rather than `int`.** `with_exposed_ports`
+  now stores `self.ports[str(port)] = None` (4.15.0 types the attribute as
+  `dict[str, Optional[int]]`), so the suite's `assert 1080 in container.ports` started failing with
+  `assert 1080 in {'1080': None}` and took five tests — and the whole `MockServer Python` pipeline,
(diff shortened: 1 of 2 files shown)


jamesdbloom:
Landed on `master` as 612b1eaf0 via rebase + fast-forward (this repo keeps master linear — no merge commits), so GitHub does not mark this PR merged. Closing.

PR #2460, opened 2026-07-26
fix(testcontainers-node): target the brace-expansion override so archiver.glob works

jamesdbloom:
## The problem

`10c3ed70e` closed CVE-2026-14257 (GHSA-mh99-v99m-4gvg) with a blanket override:

```json
"overrides": { "brace-expansion": "^5.0.8" }
```

That does resolve the tree to a single hoisted 5.0.8 and `npm audit` reports zero vulnerabilities — but **it breaks brace globbing on `master` today**.

brace-expansion 5.x changed the CommonJS export from a callable function to an object:

```js
> require('brace-expansion')
{ EXPANSION_MAX: …, EXPANSION_MAX_LENGTH: …, expand: [Function] }   // not callable
```

The minimatch copies actually installed — **3.1.5, 5.1.9 and 9.0.9** — all call it as `expand(pattern)`. Verified against master's own lockfile:

```
minimatch9 plain  *.txt       -> true
minimatch9 BRACE  *.{txt,md}  -> BROKEN: TypeError: (0 , brace_expansion_1.default) is not a function
minimatch3 plain  *.txt       -> true
minimatch3 BRACE  *.{txt,md}  -> BROKEN: TypeError: expand is not a function
minimatch5 BRACE  *.{txt,md}  -> BROKEN: TypeError: expand is not a function
```

`archiver.glob('*.{txt,md}')` crashes outright on master:

```
TypeError: expand is not a function
    at braceExpand (node_modules/readdir-glob/node_modules/minimatch/minimatch.js:119:10)
    at new ReaddirGlob (node_modules/readdir-glob/index.js:133:32)
    at Archiver.glob (node_modules/archiver/lib/core.js:752:17)
```

### Blast radius (measured, not assumed)

Narrower than it first appears: `testcontainers` copies files with `archiver.directory()` and `.append()`, which pass no brace pattern — I confirmed `directory()` still produces a valid tar on master's tree. So the container file-copy path is **not** broken. What is broken is brace globbing anywhere in this module's runtime tree, including any direct `archiver.glob()` use.

It is still worth fixing: the override silently degrades a shared dependency's contract, and it is a live trap for the next caller that uses a `{a,b}` pattern.

### Why CI didn't catch it

minimatch short-circuits patterns containing no `{`, so plain globs keep working, `npm audit` stays clean, and the unit suite stays green. Only brace patterns fail — a silent, partial breakage.

## The fix

Target the override at the two runtime consumers instead, giving them a minimatch written against the new API:

```json
"readdir-glob":    { "minimatch": "^10.2.5" },
"archiver-utils":  { "glob": { "minimatch": "^10.2.5" } }
```

`minimatch@10.2.5` depends on `brace-expansion@^5.0.5`, so both runtime copies land on the patched **5.0.8 paired with a minimatch that can actually call it**. jest keeps its own `minimatch@3.1.5` + `brace-expansion@1.1.16` pairing, untouched.

Rejected alternatives, both empirically tested:
- `"archiver": "^8.0.0"` (archiver 8 pulls patched deps) — **archiver 8 is ESM-only** and testcontainers 12 is CJS, so the package fails to load: `SyntaxError: Cannot use import statement outside a module`.
- Overriding `minimatch` globally to 10 — breaks jest's `glob@7`, which requires the v3 callable export.

## Verification

- `npm ci`, `tsc` build, **16/16 unit**, **5/5 Docker integration** (real containers)
- `npm audit --omit=dev`: **0 vulnerabilities** (8 high before, all this advisory)
- DoS actually mitigated in the installed tree: `{a,b}`×20 expands to **100,000 (bounded)** vs **1,048,576 (unbounded)** on the vulnerable version
- `archiver.glob('*.{txt,md}')` produces a valid tar again

## Regression test

New `dependency-integrity` unit test drives a brace pattern through both runtime minimatch copies and a real `archiver.glob()` tar, and asserts expansion stays bounded. Confirmed it **fails against the blanket override** (`TypeError: expand is not a function`, 2 failures), so the silent half of this cannot come back.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -7,7 +7,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
-- **The cloud blob-store, async-broker and transparent-proxy CI steps no longer OOM-kill their own build
+- **`archiver.glob()` works again in `@mockserver/testcontainers` (Node), and CVE-2026-14257 stays
+  closed.** The previous remedy for the `brace-expansion` denial of service (GHSA-mh99-v99m-4gvg,
+  patched only in 5.0.8) was a blanket `"brace-expansion": "^5.0.8"` override. That resolved the whole
+  tree to a single hoisted 5.0.8 and `npm audit` reported zero vulnerabilities — but 5.x changed the
+  CommonJS export from a callable function to an object (`{ expand, EXPANSION_MAX, ... }`), while the
+  minimatch copies actually installed (3.1.5, 5.1.9, 9.0.9) all call it as `expand(pattern)`. Every
+  glob containing a brace therefore threw `TypeError: expand is not a function`, crashing
+  `archiver.glob()` — the path `testcontainers` uses to copy files into a container. The failure was
+  invisible because minimatch short-circuits patterns with no `{`, so plain globs kept working and the
+  unit suite stayed green. The override is now targeted: `readdir-glob` and `archiver-utils`' `glob`
+  take `minimatch@^10.2.5`, which depends on `brace-expansion@^5.0.5` and is written against the new
+  API, so both runtime copies land on the patched 5.0.8 with a matching minimatch. jest keeps its own
+  `minimatch@3.1.5` + `brace-expansion@1.1.16` pairing and is untouched. `npm audit --omit=dev` still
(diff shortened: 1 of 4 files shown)


jamesdbloom:
Landed on `master` as 986cd25bb (+ changelog scoping in dbb536a4e) via rebase + fast-forward (this repo keeps master linear — no merge commits), so GitHub does not mark this PR merged. Closing.

Issue #2455, opened 2026-07-24
Jetbrains IDE Plugin 7.4.0 MockServerLLM Incorrect Request Schema


james-mchugh:
### Describe the issue

The Jetbrains IDE plugin version 7.4.0 appears to use the incorrect schema when attempting to POST LLM expectations to the MockServer of the same version.

<img width="1679" height="480" alt="Image" src="https://github.com/user-attachments/assets/4a96c6e7-22e8-421f-9192-9aed07b493b5" />

Hitting "Load into Server" results in the following error:

```
MockServer returned 400: incorrect expectation json format for: { "httpRequest" : { "method" : "POST", "path" : "/v1/chat/completions" }, "httpLlmResponse" : { "provider" : "OPEN_AI", "model" : "gpt-4o", "completion" : "Test", "finishReason" : "stop" } } schema validation errors: 3 errors: - $.httpLlmResponse.completion: string found, object expected - $.httpLlmResponse.finishReason: is not defined in the schema and the schema does not allow additional properties - $.httpLlmResponse.provider: does not have a value in the enumeration ["ANTHROPIC", "OPENAI", "OPENAI_RESPONSES", "GEMINI", "BEDROCK", "AZURE_OPENAI", "OLLAMA", "COHERE", "VOYAGE"] OpenAPI Specification: https://app.swaggerhub.com/apis/jamesdbloom/mock-server-openapi/7.4.x Documentation: https://mock-server.com/mock_server/creating_expectations.html
```

### MockServer version

7.4.0

### How are you running MockServer?

Docker

### To Reproduce

1. Spin up MockServer 7.4.0 Docker image
2. Install 7.4.0 Jetbrains IDE plugin
3. Use MockServerLLM window in IDE to send expectation to LLM
4. Observe failure

### Expected behaviour

An expectation is added to the server to mock an LLM response.

### MockServer Log

```shell

```

jamesdbloom:
Confirmed as a real bug, reproduced, and fixed in #2461 (Buildkite build 6254 green across all 13 pipelines, including the Editors pipeline that builds and tests both extensions).

### What was wrong

The tool window's builder emitted a document whose shape never existed on the server:

| Plugin sent | Server schema requires |
|---|---|
| `"provider": "OPEN_AI"` | `OPENAI` — the `Provider` enum name |
| `"completion": "Test"` (a string) | `"completion": { … }` — an object |
[...]

Issue #2450, opened 2026-07-22
Response body from file path does not work if no template engine is configured


markieww2:
### Describe the issue

Create an expectation with a body defined by an external file but WITHOUT a template engine.
Based on the description in the UI "Non (serve file verbatim)" I assumed that the file will be served 1:1 but only the filepath is returned as response, which is wrong.

<img width="313" height="158" alt="Image" src="https://github.com/user-attachments/assets/0d659450-d6d5-4053-94f7-17ee7f799e12" />

I guess you are missing the default handling of a  filebody/filebodydto with type="FILE" but no set templateType. 

**Workaround (for users):** Set the templateType even you do not need to replace content. 

Followup to https://github.com/mock-server/mockserver-monorepo/issues/2430



### What you are trying to do

Using these expectations
```
[
  {
    "httpRequest": {
      "method": "GET",
      "path": "/examples/working"
    },
    "id": "working",
    "httpResponse": {
      "statusCode": 200,
      "body": {
        "type": "FILE",
        "filePath": "/assets/examples-simple.xml",
        "templateType": "MUSTACHE",
        "contentType": "application/xml"
      }
    }
  },
  
  {
    "httpRequest": {
      "method": "GET",
      "path": "/examples/non-working"
    },
    "id": "non-working",
    "httpResponse": {
      "statusCode": 200,
      "body": {
        "type": "FILE",
        "filePath": "/assets/examples-simple.xml",
        "contentType": "application/xml"
      }
    }
  }

]
```

examples-simple.xml:
```
<tag>hello{{ request.path }}</tag>
```

### MockServer version

7.4.1-SNAPSHOT

### How are you running MockServer?

Helm / Kubernetes

### To Reproduce

Execute the requests 
```
curl http://mockserver.local/examples/non-working 
```
-> Result (wrong)
```
/assets/examples-simple.xml
```

AND

```
curl http://mockserver.local/examples/working    
```
-> Result (ok)
```
<tag>hello/examples/working</tag>
```

### Expected behaviour

Return the real payload 1:1 even there is no template engine configured

Execute the requests 
```
curl http://mockserver.local/examples/non-working 
```
-> Result (fixed, no template engine processing)
```
<tag>hello{{ request.path }}</tag>
```

AND

```
curl http://mockserver.local/examples/working    
```
-> Result (still ok)
```
<tag>hello/examples/working</tag>
```

### MockServer Log

```shell
[...]

PR #2461, opened 2026-07-26
fix: complete #2450 responseOverride truncation and #2455 JetBrains LLM expectation shape

jamesdbloom:
Resolves the two reported issues, plus one adjacent defect found while proving the first.

## #2455 — JetBrains LLM tool window sends an invalid expectation

`Load into Server` was answered with `400 incorrect expectation json format` because the builder emitted a shape that never existed server-side.

| Plugin emitted | Server schema requires |
|---|---|
| `provider: "OPEN_AI"` | `OPENAI` (enum value) |
| `completion: "Test"` (string) | `completion: { text, stopReason, usage{inputTokens,outputTokens}, streaming }` |
| top-level `finishReason`, `stream`, `usage` | not properties — `additionalProperties: false` |

**Why nothing caught it:** the builder's tests asserted its output matched the *same invented shape the builder produced* — self-derived, so they would have passed however wrong the document was. The plugin has always **bundled the correct schema** (the one MockServer validates against) and simply never checked its own output against it. It does now, and a negative case pins the validator by requiring the exact payload from the bug report to still be rejected.

The completion catalogue shared with the VS Code extension carried the same fiction (`messages`, `stream`, `toolCalls`, `finishReason`, top-level `usage`, `VERTEX_AI`) and is corrected the same way; both editors now offer the nested `completion` fields when the cursor is inside `completion`.

**Also fixed:** the `httpLlmResponse.provider` enum listed 9 of the 14 `Provider` constants, so `MISTRAL`, `XAI`, `DEEPSEEK`, `GROQ`, `OPENROUTER` were rejected with the same 400 despite each having a registered response codec. Added to the core schema, both generated editor schemas, and both copies of the OpenAPI spec, with a parity test that fails if the enum and `Provider` diverge in either direction.

## #2450 — residual `responseOverride` truncation

`6fe8b9aa3` (already on master) fixed the callback and response-template paths via `FileBodyMaterialiser`. Running the new integration test against master shows the forward **`responseOverride`** path is still wrong: the file is materialised correctly and then **truncated on the wire**, because `HttpResponse.update()` replaced the body but kept the *upstream* response's `Content-Length`. A 34-byte file behind an upstream `Content-Length: 13` arrives as 13 bytes.

Every layer above the encoder holds a correct response, which is why only a real client reading a real socket sees it. The stale header is now dropped so the encoder recomputes it; a `Content-Length` set by the override itself, and `connectionOptions.contentLengthHeaderOverride`, are still honoured, and a header-only override is untouched. This affects **every** body override, not just file bodies.

## Verification

- `mockserver-core`: **8149 + 2430** tests, 0 failures
- `FileBodyVerbatimNonStaticActionsIntegrationTest`: 4/4 over real sockets — the `responseOverride` case **fails on plain master**
- `ProviderSchemaEnumParityTest`: verified red when the schema change is reverted
- JetBrains plugin: **159** tests, 0 failures (5 of the 6 new conformance tests failed before the fix)
- VS Code: **173** tests, 0 failed
- `cargo check` clean; `OpenApiSpecSyncTest` green (it caught that only one of the two OpenAPI copies had been updated)

## Notes

- `ai_mcp_tools.html` is deliberately **not** updated: it documents the `mock_llm_failover` MCP tool's own provider enum, a separate surface whose coverage is test-backed for only 7 providers. Documenting the other five there would be an unverified claim — pre-existing drift, left for its own change.
- Master's pipeline is currently red from an unrelated Python `testcontainers` failure owned by another branch; expect that to be inherited here.


jamesdbloom:
**Buildkite build [6254](https://buildkite.com/mockserver/mockserver/builds/6254) passed — all 13 sub-pipelines green:** Java, Node, Python, Ruby, Go, .NET, Rust, PHP, Editors, Website, Infra, plus pipeline generation and client version pins.

The Editors pipeline is the one that matters for #2455 — both `VS Code extension compile + test` and `JetBrains plugin tests` passed on the agents, not just locally.

Rebases cleanly onto current master (`260cf369e`); GitHub reports the PR `MERGEABLE` / `CLEAN`.

Verification recap:
- `mockserver-core`: 8149 + 2430 tests, 0 failures
- `FileBodyVerbatimNonStaticActionsIntegrationTest`: 4/4 over real sockets — the `responseOverride` case **fails on plain master**, which is what makes it load-bearing
- `ProviderSchemaEnumParityTest`: verified red when the schema change is reverted
- JetBrains 159 tests / VS Code 173 tests, 0 failures (5 of 6 new conformance tests failed before the fix)

Left as draft for a maintainer to merge — I don't push to master.

PR #2715, opened 2026-09-19
DO NOT MERGE — item 16 PR-path control: deliberate allocation regression

jamesdbloom:
## Do not merge — this is a deliberate regression

This PR exists to be observed failing, then closed. It is the executed negative control for
**item 16 (PR path)** of the performance programme.

### What it does
Allocates ~16 KB per decoded request in `FullHttpRequestToMockServerHttpRequest`, escaping into a
static volatile sink so the JIT cannot scalar-replace it away.

`InboundDecodeBenchmark` measures **37,544 B/op** against a committed floor of **47,000** — 9,456 B/op
of headroom. A 16,384-byte array per op clears that decisively.

### Why it exists
The gate has already been shown to catch a regression on `master` (build 2262) — but that is
*post-merge detection*. The programme records that **"merge-blocking is a property of wiring, not of a
check"**, and that asserting it without executing it is an inference this programme has been burned by
once already.

So the claim under test is narrower and load-bearing: the `:scales: per-merge allocation gate` step in
`.buildkite/pipeline-java.yml` is wired **unconditionally**, before the `master`-only `wait`, and a
failure there propagates through `trigger-pipeline.sh` (which exits 1 on a failed downstream build) to
red the orchestrator build and block this PR **before** merge.

### Expected outcome
`:scales: per-merge allocation gate (item 16)` fails, naming `InboundDecodeBenchmark` over its floor,
and this PR is blocked. Then it gets closed and the branch deleted.


jamesdbloom:
Control executed and recorded — closing as intended.

The per-merge allocation gate failed on the **PR** build (java 2310), pre-merge:

```
:x: InboundDecodeBenchmark   alloc=53968 B/op  floor=47000
✅ MatchingBenchmark          alloc=1498011     floor=1850000
✅ MatchingBenchmark_detailed alloc=2096357     floor=2650000
✅ ResponseWriteBenchmark     alloc=36537       floor=46000
```

The load-bearing part is what **passed**: `:maven: build` and the dashboard gate were both green, so the allocation gate was the *sole* blocker. "The PR was blocked" and "the gate blocked the PR" are different claims, and only the second is the control.

Recorded against item 16 in the performance programme (`c968e86dc`). This branch was never merged and is now deleted.

